Issue #6, opened 2020-04-25
Multiple streams


scullabyte:
How would you implement multiple streams with the `StreamViewModel` currently? I would imagine combining and transforming streams to get multiple types out of them could get tedious. 

#### My two cents:

I would think the best implementation would be the way you set up Provider previously:
```dart
providers = [
    ..._independentServices,
    ..._dependentServices,
    ..._consumableServices,
  ];
```

Maybe it would it make sense to have something like this:

```dart
abstract class StreamsViewModel extends BaseViewModel {

List<StreamsMixin<T>> get streams;

StreamsViewModel(){
    _listenToStreams(streams);
}
  void _listenToStreams(List<StreamsMixin> streams) {
    _streams = streams;
    for (var stream in _streams) {
      stream.addListener(_indicateChange);
    }
  }
```

Where the `StreamsMixin` does the same thing that the `StreamViewModel` does. Does that make sense? If I'm missing something and this is already doable, then an example of multiple streams would go a long way. Thanks!

FilledStacks:
Hey @scullabyte this is not possible yet. The only thing that I struggle with when it comes to multiple streams is how we'll distinguish the data that's fetched and how it will be stored. Currently it's available through the data property but if there's multiple streams what will those be stored in?

I was thinking of using a map with keys or an enum for readability so you can access the streams using that. So you'll register

```
Map<enum, Stream<T>> get streamsMap = {
  AuthenticationStatus: _authService.authenticationStatusStream,
  Orders: _orderStervice.ordersStream,
}

// Accessing a stream
var latestOrders = data[Orders]
```

I haven't tried it out but I am thinking about it, the same with the Future viewmodel. Multiple Futures and cancel / timeout / retry functionality built into it. I'm writing today, hoping I can get the written done and the video in one day for the new tutorial series episode 1. I'll be improving this package as we're developing a new food delivery service customer app so It'll get constant updates and fixes over the next few weeks.

FilledStacks:
Pushed base implementation. We'll refactor as we find better ways of doing things or the requirements change. Thanks for all the effort

PR #9, opened 2020-04-27
Multi stream

scullabyte:
This is a basic implementation of a multi-stream solution outlined here #6 . I left a lot of implementation open so you could change it however you'd like. This basically wraps a version of your `StreamsViewModel` class into a reusable cached stream factory so that you can keep the same naming conventions and streams across multiple files while not hogging up more memory. This lets you transform data on the streams just as you would with the single-stream implementation.

None of this has been tested in an emulator yet, it only works in the complier.


FilledStacks:
@scullabyte that's exactly what I was thinking. Minus the caching part. Thanks for the implementation. I'll definitely look at this when I implement it. Gonna have to test with a few scenarios first before getting it in here.

scullabyte:
Ok, so I have been tinkering with it with my actual emulated apps and made a few changes to get it to work. Just pushed the commit.

I've changed it so it doesn't use a factory or cache system because it was more trouble than it was worth. And I found that for some reason the ChangeNotifier wasn't working as planned, so I added the hack below to get it to work after building the streams. I'm sure there's a better way to notify on load though.

```dart
abstract class MultiStreamViewModel extends BaseViewModel {
  Map<String, Stream> get streams;
  Map<String, StreamData> streamsOut = {};

  void buildStreams() async {
    streams.forEach((name, stream) {
      streamsOut[name] = StreamData(stream: stream);
      streamsOut[name].listen();
    });
    // TODO: Maybe fix this hack?
    await Future.delayed(Duration(microseconds: 1));
    // For some reason the ChangeNotifier wasn't updating in my emulator
    // So had notify listeners after a delay.
    notifyListeners();
  }
}
```

lib/_base_viewmodels.dart
```diff
@@ -115,24 +104,29 @@ class StackedStream<T> extends SingleDataSourceViewModel<T> {
...
+  @override
+  void dispose() {
+    _streamSubscription.cancel();
+    onCancel();
+    super.dispose();
+  }
 }
 
-abstract class StreamsViewModel extends BaseViewModel {
+abstract class MultiStreamViewModel extends BaseViewModel {
```
FilledStacks: This was a change that I was going to ask for after reading the first set of changes. I like the naming updates

lib/_base_viewmodels.dart
```diff
@@ -115,24 +104,29 @@ class StackedStream<T> extends SingleDataSourceViewModel<T> {
...
-  @override
-  void dispose() {
-    streamsOut.forEach((_, stream) {
-      stream.streamSubscription.cancel();
+      streamsOut[name] = StreamData(stream: stream);
+      streamsOut[name].listen();
     });
-    super.dispose();
+    // TODO: Maybe fix this hack?
+    await Future.delayed(Duration(microseconds: 1));
```
FilledStacks: I don't think we need notifyListeners here. There's nothing that will change after the subscription. We should be firing off the life cycle event here like in StreamViewModel. onSubscribe, the other life cycle events.

FilledStacks (changes requested):
There are a few things that has to be added. 

- **Add lifecycle events**: All the life cycle events should still be called per stream and pass their key back for reference. 
- **Removal of duplicate code**: The code in .listen should be a shared function to avoid repeating all of that code twice. It could be moved into a private class that contains the lifecycle abstraction and the general listen, subscribe and unsubscribe functionality. 

Doing that should remove a lot of the duplicate code currently in the implementation.

scullabyte:
Ok, so the problem I'm having with making a private class for the `StreamData.listen` is that the `notifyListeners` inside the private class aren't updating the viewModel. If I debug the `StreamData.listen` method I can see that the data is coming in, however, the data never makes its way back to the `streamData['item'].data` on the actual ViewModel. It's late here, so I know I'm missing something obvious. If you can spot what I did wrong on this lmk and I'll change it in the morning.

FilledStacks:
Aaaaaah, I see. I'll swap to your fork now now and take a look. I'm finishing up and pushing the MultipleFuturesViewModel now. Just writing some basic tests because I know this implementation will change quite a lot in the coming weeks as I use the package in production

FilledStacks:
It's because you're calling notifyListeners for SingleDataSourceViewModel inside the StreamData but the MultiStreamViewModel doesn't inherit from StreamData so it won't notify. There's two concepts mixed up there.

I appreciate the implementation But I think your previous version was better. I'm not paying you to do any of this so I'll take it off your hands haha. I have an hour or so free coming up so I'll implement it and push.

Thanks for all the effort.

FilledStacks:
Turns out I couldn't get to it. Pull in the latest changes and look at my Futures implementation. You can make use of the _multiDataSource class that's there and implement the same thing for streams as well. Remove the StreamData construct. It's a bit confusing at the moment. I've started already so you can leave it as well. I'l only be able to get to it tomorrow morning.

scullabyte:
Awesome man. I just started looking through it and I'll start tinkering with it in the next hour or so!

scullabyte:
Just FYI, I added all the tests for `StreamViewModel` and `MultipleStreamViewModel`. I had to change `StreamViewModel` to handle/set busy. Only problem I see is that if the user overrides the onData/onCancel/onError calls, they will have to implement their own setbusy calls.

scullabyte:
I think I'm going to handle the individual stream 'busy' status by utilizing your `runBusyFuture` method, where all stream data will be initially handled as a future until the first stream data is received. So this would have a completer running onData, onError and onCancel. Lmk if you were thinking of something different.

lib/_base_viewmodels.dart
```diff
@@ -174,29 +181,46 @@ abstract class MultipleFutureViewModel extends _MultiDataSourceViewModel {
...
+  StreamData streamData;
+  get data => streamData.data;
+  get dataReady => streamData.dataReady;
+  get hasError => streamData.hasError;
+  get isBusy => streamData.isBusy;
+
+  //TODO: Integrate read/write for onData, onCancel, onSubscribed, onError and transformData properties
+  //TODO: Figure out a more elegant way of integrating StreamData
+  //TODO: Get this to work with generic stream types again
+  //TODO: Possibly have a default setbusy for overwrittenn lifecycle events
```
scullabyte: For now was thinking about just having setters override the lifecycle events on the `StreamData` class. Let me know if you can think of a better way to implement this, possibly with generics again.

lib/_base_viewmodels.dart
```diff
@@ -66,63 +82,209 @@ abstract class ReactiveViewModel extends BaseViewModel {
...
+    onSubscribed();
   }
 
   /// Called when the new data arrives
   ///
   /// notifyListeners is called before this so no need to call in here unless you're
   /// running additional logic and setting a separate value.
-  void onData(T data) {}
+  void onData(T data) {
+    setBusy(false);
```
FilledStacks: We don't want any automatic busy states for streams. If you have multiple streams and one fails the model will be set to not busy automatically which might not be what the dev is expecting. For now we can remove all setBusy calls

lib/_base_viewmodels.dart
```diff
@@ -66,63 +82,209 @@ abstract class ReactiveViewModel extends BaseViewModel {
...
+  //TODO: Figure out a more elegant way of integrating StreamData
+  //TODO: Get this to work with generic stream types again
+  //TODO: Possibly have a default setbusy for overwrittenn lifecycle events
+
+  Future initialise() async {
+    streamData = setupStream(stream);
+  }
+}
+
+class StreamData<T> extends _SingleDataSourceViewModel<T> {
```
FilledStacks, line 292: We should make this class private so it's not available for use outside of this file

FilledStacks (changes requested):
Small changes.
- no automatic setBusy's
- private StreamData class.

Thanks a lot man. Appreciate you helping out. Saves me some time.

scullabyte:
Any thoughts on how to override the lifecycle events for individual streams inside the ViewModel now? If we can still do it with current implementation lmk, otherwise just let me know what to change and I'll get it working and tested.

scullabyte:
Ok man.. I'm pretty proud of this last update. I'm actually using this on my own app while I've been writing this and it solves a lot of problems for me. We have multiple streams being able to be set with their lifecycles able to be altered independently. 

I had to make the `StreamData` class public because we're using it as a getter in the `MultipleStreamViewModel` (instead of `StreamDataMap`). So the `StreamData` now takes a stream, then optional parameters for each of the lifecycle events you want to override (defaulting to the MultipleStreamViewModel lifecycle events). 

```dart
  // The only thing you need to override
  @override
  Map<String, StreamData> get streamDataMap => {
	  _TextStream: StreamData(textStream()),
      _NumberStream: StreamData(
        numberStream(),
        onData: _loadedNumbers, // <--- this is optional
      )
	};
	  void _loadedNumbers(data) {
		// Handles single stream 
	    loadedData = data;
	  }
```

I fixed everything that I had broken your on original implementation (with lifecycle events). Let me know what you think!

FilledStacks:
I like it! and the fact that you updated the unit tests to match. I'll merge it in and get the update up. With all these things I'll see how it feels while developing multiple code bases and it'll grow / shrink depending on the core functionalities that are required. 

Thanks for the contribution. I'll merge this in now and publish a new update.

FilledStacks:
This actually isn't working. When calling the dataReady function on the `_MultiDataSourceViewModel` it throws an exception. I'm gonna have to change your implementation to line up with the multi future viewmodel for consistency. The more I read it the more things I see that can be simplified. I'll get to it when I have some time.

PR #10, opened 2020-04-27
Add context for viewModelBuilder

necatisozer:
`BuildContext` is needed for `viewModelBuilder` function thus we may call `Provider` or other context needed dependencies here.

Example usage: 
```dart
class BuilderWidgetExampleView extends ViewModelBuilderWidget<HomeViewModel> {

  @override
  bool get reactive => false;

  @override
  bool get createNewModelOnInsert => false;

  @override
  bool get disposeViewModel => true;

  @override
  Widget builder(
    BuildContext context,
    HomeViewModel model,
    Widget child,
  ) {
    return Scaffold(
      body: Center(
        child: Text(model.title),
      ),
      floatingActionButton: FloatingActionButton(
        onPressed: () => model.updateTitle(),
      ),
    );
  }

  @override
  HomeViewModel viewModelBuilder(BuildContext context) {
    final dependency = Provider.of<Dependency>(context);
    return HomeViewModel(dependency);
  }
```

### DIFF
--- a/lib/_viewmodel_builder_widget.dart
+++ b/lib/_viewmodel_builder_widget.dart
@@ -1,4 +1,4 @@
-import 'package:flutter/cupertino.dart';
+import 'package:flutter/widgets.dart';
 
 import '_viewmodel_builder.dart';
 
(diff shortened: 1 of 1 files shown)


FilledStacks:
Oh yes, thank you. I forgot about the pure provider users. Thanks for the update. Merging in now

FilledStacks:
I might have to make it optional to avoid breakages

PR #30, opened 2020-05-05
[stacked] Add const constructors

### LINKED ISSUE #25: [stacked] Const constructors for view model widgets
Why be there no `const` constructors for the immutable [`ViewModelWidget` class](https://github.com/FilledStacks/stacked/blob/80bcaaddf51237c0870093a44b634e4a4b1cd590/lib/_viewmodel_widget.dart#L6)?

I feel like https://github.com/FilledStacks/stacked/blob/80bcaaddf51237c0870093a44b634e4a4b1cd590/lib/_viewmodel_widget.dart#L9

Should actually look like this:

```dart
const ViewModelWidget({Key key, this.reactive = true}) : super(key: key);
```

What is the rationale behind that decision?

creativecreatorormaybenot:
Resolves #25.

### Miscellaneous

I added a https://shields.io badge for the Pub version in order to be able to see the currently uploaded Pub version from the GitHub repo and also because it allows to go to the Pub package page with a single click. This way you can easily navigate from Pub to GitHub (using the "Repository (GitHub)" button) and now also the other way around.  
Remove it if you think it should not be included.

Sidenote: I was wondering why the changelog uses headers to show changes. I followed it now because I am fine with doing so. However, I find it a bit hard to read the changelog when the text is so big that it spans across the entire screen.  
I like the [way the Flutter team does it](https://github.com/flutter/plugins/blob/master/packages/url_launcher/url_launcher/CHANGELOG.md). This has nothing to do with the PR; though, was just wondering 😅

### DIFF
--- a/packages/stacked/CHANGELOG.md
+++ b/packages/stacked/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.4.1 - 2020-05-05 - Added const constructors
+
+- Added the `const` keyword to the following widget constructors:
+  `ViewModelBuilder.nonReactive`, `ViewModelBuilder.reactive`, and `ViewModelWidget`
+
 ## 1.4.0 - 2020-05-04 - [Breaking] Moved private files into src folder
 
 - Fix imports, import visibility and changed file names.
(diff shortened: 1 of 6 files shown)


FilledStacks:
Thanks man. I appreciate all of the efforts for some of the maintenance things.

I also like the way the Flutter team does it haha. I follow the template that's in the flutter create template that gets generated. But now that I look at it, it might be old and not a great way of doing things. 

I will update all the packages in my break after hitting the gym. Will push the const update as well.

PR #40, opened 2020-05-13
Modifying Dialog Service

YazeedAlKhalaf:
This PR has the following modifications:

• Modified the dialog service so that it can handle platform choosing for dialog designs 👍, more info in the file where it is documented 📜✨
• Modified example app to show how the dialog service works now. 🚀

### DIFF
--- a/packages/stacked_services/.gitignore
+++ b/packages/stacked_services/.gitignore
@@ -18,7 +18,7 @@
 # The .vscode folder contains launch configuration and tasks you configure in
 # VS Code which you may wish to be included in version control, so this line
 # is commented out by default.
-#.vscode/
+.vscode/
 
 # Flutter/Dart/Pub related
 **/doc/api/
--- a/packages/stacked_services/example/.gitignore
+++ b/packages/stacked_services/example/.gitignore
@@ -18,7 +18,7 @@
 # The .vscode folder contains launch configuration and tasks you configure in
 # VS Code which you may wish to be included in version control, so this line
 # is commented out by default.
-#.vscode/
+.vscode/
 
 # Flutter/Dart/Pub related
 **/doc/api/
--- a/packages/stacked_services/example/lib/ui/views/home_screen.dart
+++ b/packages/stacked_services/example/lib/ui/views/home_screen.dart
@@ -56,10 +56,9 @@ class _HomeScreenState extends State<HomeScreen> {
             ),
             OutlineButton(
               onPressed: () async {
-                await _dialogService.showDialog(
+                await _dialogService.showConfirmationDialog(
                   title: 'Test Confirmation Dialog Title',
                   description: 'Test Confirmation Dialog Description',
-                  isConfirmationDialog: true,
                 );
               },
               child: Text(
--- a/packages/stacked_services/example/lib/ui/views/home_screen.dart
+++ b/packages/stacked_services/example/lib/ui/views/home_screen.dart
@@ -30,7 +30,7 @@ class _HomeScreenState extends State<HomeScreen> {
           mainAxisAlignment: MainAxisAlignment.center,
           children: <Widget>[
             Text(
-              'Press the button below to show a dialog',
+              'Press the button below to show a regular dialog',
               softWrap: true,
               style: TextStyle(
                 fontSize: 14,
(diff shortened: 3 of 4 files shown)


YazeedAlKhalaf:
@FilledStacks

BrennenRocks:
Ability to change the text color will help with Dark Mode

YazeedAlKhalaf:
@BrennenRocks yup you are right, that is why I did allow the developer to change the whole TextStyle if they choose to. Else it will be black. You can check the code

YazeedAlKhalaf:
I would like to see if you guys like the change I made to the `showConfirmationDialog` by removing it and making it a part of the `showDialog` function.

It made the file a lot shorter and less code with the same implementation, it will break the API but I think it is not a big deal now since this is still in the `0.y.z` phase of development 😊.

I believe that making the code shorter is better 😁

FilledStacks:
Thanks for the PR. I appreciate it :) 

Shorter code is not always "Better". "Better" code depends on what you value personally in a code base. I like to have domain language. For instance I'd like to say "Call showConfirmation Dialog and check the confirmed action" instead of saying "Call showDialog and set showConfirmation Dialog to true". They both also read different. 

```
var confirmationResponse = _dialogService.showConfirmationDialog()
```
vs
```
var confirmationResponse = _dialogService.showDialog(confirmation: true)
```

They're both perfectly fine but if there's more values in the function call, which there will be it's harder to see if it's a confirmation dialog. The other part of the argument is that because there's no function overloads it's hard to establish an additional signature, which is something I often used in C#. 

If you have two different functions you can make different things required for that function. The only thing I'd change in the PR is removing the new confirmation dialog functionality and make sure it still works like the old one. Otherwise I'm happy with the changes.

YazeedAlKhalaf:
@FilledStacks 
Okay I will be changing it today and return to using two functions 🥳

But I still prefer to do it in one function because the thing that is making a difference between a confirmation dialog and a regular dialog is the `cancelText` so if the user adds the `cancelText`, he will get a confirmation dialog although he chooses to get a regular dialog

YazeedAlKhalaf:
Made the changes, I guess it is ready to be merged 😊✨🚀

FilledStacks:
One thing I just noticed is that you expect the TextStyle to be passed in from the ViewModels. The ViewModels should have none of the UI code to keep the separation of concerns clear when developing an app. Those have to be removed from the function signature. Set a default text style for now we'll create a separate config where we can set the styles of the dialog. 

Don't do that in this PR. Remove the TextStyles from the Function signature and use default styles. ViewModels shouldn't know anything about the UI. That's why we provide that abstraction. So the ViewModel code can contain logic only.

YazeedAlKhalaf:
@FilledStacks  There is a default textstyle but you can provide a custom one if you feel that it is not suiting the needs.m

Ok, one question: so if I don’t change it now, when?

FilledStacks:
Once the functionality for this PR is confirmed and merged. If you keep adding more stuff the feature will never get done. One thing to do as you grow your software development career is to stick to the task at hand, complete it 100%, don't add anything you think might be great, or anything you would like to see until the current functionality at hand it complete. Take note of everything you're thinking of while busy, write it down, when you're done you assess and then see if it's still a good idea to add those things. 

I've had features from developers drag on for weeks and the original functionality would have taken maybe a day to do at max. Stick to the spec, deliver, then improve. Same principles that's followed during TDD, Red, Green, REFACTOR. You have a problem - RED (in this case, no platform specific dialogs), Fix the problem - GREEN (Add basic implementation, no styling, just get the solution to the problem), Improve on the new functionality - REFACTOR (Add a separate DialogStyleConfig class where the styles are pulled from, this is the class the user will setup before the app is run in main where they can supply a const style for their dialogs). 

That refactor step is a separate PR where we focus ONLY on making the feature better, you can only get there once the feature has been implemented. I hope that makes sense.

YazeedAlKhalaf:
Ah okay, so I don't change anything now. How much time does it take to finish your review?

FilledStacks:
You need to do what I mentioned above. 

> One thing I just noticed is that you expect the TextStyle to be passed in from the ViewModels. The ViewModels should have none of the UI code to keep the separation of concerns clear when developing an app. Those have to be removed from the function signature. 

The other thing we'll do in a different PR. But the TextStyles has to be out of the signature. Set the default text styles only for now. That should be the last change.

YazeedAlKhalaf:
Okay, Got you 👌😊

FilledStacks:
Your pull request has been merged in. I need to implement some basic things to remove the code duplcation as well as the nested function code then I'll push it up and you can see the reviewed code for the implementation.

YazeedAlKhalaf:
Okay, that’s great to hear 👂! 
Btw, like your work 😉

PR #43, opened 2020-05-16
Added support to Flutter_web and fix Get version

jonataslaw:
The project used Dart: io which made it incompatible with Flutter_web.
The Get version was updated in pubspec to correct the error with null Get.key.


YazeedAlKhalaf:
This is great man 👍👍

FilledStacks:
Dopeness @jonataslaw . Thanks for that fix.

PR #54, opened 2020-05-26
Implementing Open/Closed Principle 

ParthaGudivada:
1. Fulfilled the TODO to perform the actions for specialized view models.
2. Defined the abstract class that defines what action should be performed
3. Implement the interface by the classes that need to perform the specialized actions.

### DIFF
--- a/packages/stacked/lib/src/base_view_models.dart
+++ b/packages/stacked/lib/src/base_view_models.dart
@@ -139,7 +139,8 @@ class _MultiDataSourceViewModel extends DynamicSourceViewModel {
 }
 
 /// Provides functionality for a ViewModel that's sole purpose it is to fetch data using a [Future]
-abstract class FutureViewModel<T> extends _SingleDataSourceViewModel<T> {
+abstract class FutureViewModel<T> extends _SingleDataSourceViewModel<T>
+    implements IAdditionalSetup {
   /// The future that fetches the data and sets the view to busy
   @Deprecated('Use the futureToRun function')
   Future<T> get future => null;
@@ -178,10 +179,15 @@ abstract class FutureViewModel<T> extends _SingleDataSourceViewModel<T> {
 
   /// Called after the data has been set
   void onData(T data) {}
+
+  doSetup() {
+    runFuture();
+  }
 }
 
 /// Provides functionality for a ViewModel to run and fetch data using multiple future
-abstract class MultipleFutureViewModel extends _MultiDataSourceViewModel {
+abstract class MultipleFutureViewModel extends _MultiDataSourceViewModel
+    implements IAdditionalSetup {
   Map<String, Future Function()> get futuresMap;
 
   Completer _futuresCompleter;
@@ -241,10 +247,15 @@ abstract class MultipleFutureViewModel extends _MultiDataSourceViewModel {
   void onError({String key, error}) {}
 
   void onData(String key) {}
+
+  doSetup() {
+    runFutures();
+  }
 }
 
 /// Provides functionality for a ViewModel to run and fetch data using multiple streams
-abstract class MultipleStreamViewModel extends _MultiDataSourceViewModel {
+abstract class MultipleStreamViewModel extends _MultiDataSourceViewModel
+    implements IAdditionalSetup {
   // Every MultipleStreamViewModel must override streamDataMap
   // StreamData requires a stream, but lifecycle events are optional
   // if a lifecyle event isn't defined we use the default ones here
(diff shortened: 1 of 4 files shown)


ParthaGudivada:
Ran the unit tests and `futureviewmoodel_test` failing (failing on the `master`. for the same reason).  Based on the last PR, it is still being worked upon to handle the error.

FilledStacks:
@ParthaGudivada Thank you for submitting a PR! I appreciate it. Some feedback below

Futures that run in void functions does not behave properly. A future has to run in a function that returns a Future to allow for a Future that throws an exception to be properly handled. 

I think the function should be initialise instead of doSetup since it covers setup and running initialise logic like the futures. Make the function return a Future.

As for the tests, this package runs on stable so you'll have to ensure all the unit tests pass that exists and update them to call the new function instead of the old one.

ParthaGudivada:
Thanks for the feedback.  Working on that.

ParthaGudivada:
I am getting errors in executing unit test for the future.  Please take a look at the [PR](https://github.com/FilledStacks/stacked/pull/55).

ParthaGudivada:
> I think the function should be initialise instead of doSetup since it covers setup and running initialise logic like the futures. Make the function return a Future.

```flutter
void _initialiseSpecialViewModels() {
    // Add any additional actions here for spcialised ViewModels
     if (_model is FutureViewModel) {
      (_model as FutureViewModel).runFuture();
    }

    if (_model is MultipleFutureViewModel) {
      (_model as MultipleFutureViewModel).runFutures();
    }

    if (_model is StreamViewModel) {
      (_model as StreamViewModel).initialise();
    }

    if (_model is MultipleStreamViewModel) {
      (_model as MultipleStreamViewModel).initialise();
    }
  }
```

Since the existing method `_initialiseSpecialViewModels()` handles `initialise` and `runFutures`, it might not benefit to have return type of `Future`.  Your suggestion of renaming `doSetup` to `initialise` simplifies the code further.  I currently updated the PR on this.

packages/stacked/lib/src/base_view_models.dart
```diff
@@ -498,3 +509,8 @@ class StreamData<T> extends _SingleDataSourceViewModel<T> {
     super.dispose();
   }
 }
+
+/// Interface: Additional actions that should be implemented by spcialised ViewModels
+abstract class IAdditionalSetup {
```
FilledStacks: None of the interfaces use the 'I' naming convention. Please change this to Initialisable instead of Additional setup. This will align the name of the function and the interface name.

packages/stacked/lib/src/base_view_models.dart
```diff
@@ -238,10 +244,15 @@ abstract class MultipleFutureViewModel extends _MultiDataSourceViewModel {
   void onError({String key, error}) {}
 
   void onData(String key) {}
+
+  initialise() {
```
FilledStacks: Instead of having a function that calls the runFutures function. Change runFutures to be the initialise function from the Initialisable interface.

packages/stacked/lib/src/base_view_models.dart
```diff
@@ -175,10 +176,15 @@ abstract class FutureViewModel<T> extends _SingleDataSourceViewModel<T> {
 
   /// Called after the data has been set
   void onData(T data) {}
+
+  initialise() {
```
FilledStacks: Instead of calling runFuture from the initialise function we can instead change the runFuture function to be called initialise.

FilledStacks (changes requested):
With the new code I pushed the unit tests should be running again. Please ensure that they all pass and that they're updated to match the new code changes.

ParthaGudivada:
@FilledStacks I did not get notification on this.  Will work on the feedback.

packages/stacked/lib/src/base_view_models.dart
pgudivadaDiligent: Renamed the interface and updated the same for the classes that implements this interface.

packages/stacked/lib/src/base_view_models.dart
pgudivadaDiligent: Renamed the function.

packages/stacked/lib/src/base_view_models.dart
pgudivadaDiligent: Renamed the function.

pgudivadaDiligent:
@FilledStacks Renamed the interface as suggested and all unit tests ran successfully.

FilledStacks:
Awesome @ParthaGudivada I will merge this in and it'll be pushed out with the next set of updates.

PR #55, opened 2020-05-27
Error with unit test

ParthaGudivada:
1. Throw error when there is an issue with the execution of future rathe…r than returning null
2. Updated the unit test so that the test view model  does not have to throw error as the basemodel itself throws the error

@FilledStacks For me unit tests are failing for future test `When a future fails it should indicate there\'s an error and no data` as the `runBusyFuture` is return `null` for the said error condition.


FilledStacks:
Hey, I reverted the decision to return null for busyFuture errors. Don't have more time for this today. I'll look at it again tomorrow.

PR #71, opened 2020-06-13
Update dialog_service.dart

### LINKED ISSUE #70: Missing parameter in `showCustomDialog` method
The `DialogRequest` is supposed to have a parameter `customData` (which it has) but it seems there is no way to pass it in the `showCustomDialog` method.

This is the version of `StackedServices` I'm using: 
 `stacked_services: ^0.4.0`

fribeau:
Resolves #70


RodolfoSilva (approved):
LGTM!

FilledStacks:
Thanks. Will push with some other fixes in a bit.

PR #75, opened 2020-06-14
Make the showCustomSnackBar callable inside initiState like showSnackbar.

RodolfoSilva:
This will prevent this common exception: 
```
Unhandled exception: setState() or markNeedsBuild() called during build.
```

The original implementation use the same technique:

https://github.com/jonataslaw/get/blob/5ccbb96da82669472847aaf071b00e47a0bb641c/lib/src/get_main.dart#L481

https://github.com/jonataslaw/get/blob/5ccbb96da82669472847aaf071b00e47a0bb641c/lib/src/get_main.dart#L563-L570


FilledStacks:
awesome update. Thanks

PR #84, opened 2020-06-21
Added required tag for necessary arguments without which an error might be given

nikunjdk:
I am a beginner in flutter and have started following your architecture (Love it!). But one problem I faced is that when using functions not used in the tutorial I am not sure what all arguments are necessary to pass and what all are optional. I believe that adding @required tag is an easy way of solving this problem for many people. Since I am new to it I may have missed/added the tag in some necessary/unnecessary places. Please check it once. Thanks!

### DIFF
--- a/packages/stacked_services/lib/src/navigation_service.dart
+++ b/packages/stacked_services/lib/src/navigation_service.dart
@@ -88,7 +88,7 @@ class NavigationService {
   /// - rightToLeftWithFade
   /// - leftToRightWithFade
   /// - cupertino
-  Future<dynamic> navigateWithTransition(Widget page,
+  Future<dynamic> navigateWithTransition(@required Widget page,
       {bool opaque, String transition, Duration duration, bool popGesture}) {
     return Get.to(
       page,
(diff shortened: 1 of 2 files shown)


FilledStacks:
Hey there,

I don't know if you even used those functions in the navigation service. A positional argument is required. You'll get a compiler error if you don't add that. Instead of adding the same comment to all the lines of code where you use it you can remove all the @required tags from the positional arguments. It serves no purpose. The required tag is to mark a named / option parameter as required. Positional arguments are always required, your code won't run without it.

nikunjdk:
Ohh ok got it, my bad😅.

nikunjdk:
Actually I was getting a run time error when I only passed the title argument and not the message argument in snack bar service. So I thought that making the message argument required is a good way to prevent such accidents for anybody in the future.

FilledStacks:
@nikunjdk for the Snackbar service that should be fine.

nikunjdk:
Cool, then should I close this or will you do it?

FilledStacks:
You can remove all the other required annotations from the navigation service and and make another PR. 

It title also required in the snackbar service?

nikunjdk:
No, only message is required in the snackbar service I guess. It does not give an error if we just provide a message without giving the title.

PR #90, opened 2020-06-26
📚 Add Documentation for Stacked Services!

YazeedAlKhalaf:
This pull request contains the following changes:
- Add New Documentation for `NavigationService` and `SnackbarService`.
- Some typos fixed.

The part of custom Snackbar is not written yet, because there is a todo about getting UI out like what you did with the dialog service. That's why I have left it. but whenever you finish the `showCustomSnackbar` and it is stable and no todos there, I will add the docs!

I hope it is helpful!


YazeedAlKhalaf:
@FilledStacks

FilledStacks:
@YazeedAlKhalaf Please add the a section that tells the dev that they have to set the navigation key on their material / cupertino app. They can do that using the key on the dialog service if they only want to use the dialog service or they can do that using the key on the navigation service. 

The key only has to be set once for it to work for all three services.

FilledStacks:
I'll merge it in and fix it up now. I have some time

YazeedAlKhalaf:
Thanks 💙

PR #110, opened 2020-07-27
Expose barrierDismissible for confirmationDialog

sebastianbuechler:
Added the parameter barrierDismissible  for exposure in the confirmationDialog and adjusted the example to use the new functionality directly.

### DIFF
--- a/packages/stacked_services/example/lib/ui/views/home_screen.dart
+++ b/packages/stacked_services/example/lib/ui/views/home_screen.dart
@@ -92,7 +92,7 @@ class _HomeScreenState extends State<HomeScreen> {
                 await _dialogService.showConfirmationDialog(
                   title: 'Test Confirmation Dialog Title',
                   description: 'Test Confirmation Dialog Description',
-                  barrierDismissible = true,
+                  barrierDismissible: true,
                 );
               },
               child: Text(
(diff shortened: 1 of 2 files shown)


packages/stacked_services/example/lib/ui/views/home_screen.dart
```diff
@@ -92,6 +92,7 @@ class _HomeScreenState extends State<HomeScreen> {
                 await _dialogService.showConfirmationDialog(
                   title: 'Test Confirmation Dialog Title',
                   description: 'Test Confirmation Dialog Description',
+                  barrierDismissible = true,
```
FilledStacks: This doesn't compile

FilledStacks (changes requested):
example code doesn't compile

packages/stacked_services/example/lib/ui/views/home_screen.dart
sebastianbuechler: Mixed up the colon notation with equal signs in two places.

FilledStacks:
Thanks. Will push in new update.

FilledStacks:
it's in 0.4.11

sebastianbuechler:
Thank you

PR #111, opened 2020-07-27
Fixed routes to breaking changes in auto_routes package

sebastianbuechler:
Since version 0.6.0 the auto_route package requires a list instead of class fields for routes declaration.
See: [https://stackoverflow.com/questions/62853598/flutter-auto-route-generator-not-building-formatexception-not-an-instance-of-l](https://stackoverflow.com/questions/62853598/flutter-auto-route-generator-not-building-formatexception-not-an-instance-of-l)

Not sure if the newly generated routes.gr.dart file needs to be included or not.

### DIFF
--- a/packages/stacked_services/example/lib/app/routes.dart
+++ b/packages/stacked_services/example/lib/app/routes.dart
@@ -7,12 +7,22 @@ import '../ui/views/second_screen.dart';
 
 // Defining routes and global transitions
 @CustomAutoRouter(
-    transitionsBuilder: TransitionsBuilders.zoomIn, durationInMilliseconds: 400)
-class $Router {
-  @initial
-  HomeScreen homeScreenRoute;
-
-  FirstScreen firstScreenRoute;
-
-  SecondScreen secondScreenRoute;
-}
+    routes: <AutoRoute>[
+    MaterialRoute(
+      page: HomeScreen,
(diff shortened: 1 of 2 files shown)


packages/stacked_services/example/lib/app/routes.dart
```diff
@@ -7,12 +7,22 @@ import '../ui/views/second_screen.dart';
...
-    transitionsBuilder: TransitionsBuilders.zoomIn, durationInMilliseconds: 400)
-class $Router {
-  @initial
-  HomeScreen homeScreenRoute;
-
-  FirstScreen firstScreenRoute;
-
-  SecondScreen secondScreenRoute;
-}
+    routes: <AutoRoute>[
```
RodolfoSilva: Hey @sebastianbuechler great job.
Just a small change, can you adjust the indentation in the `routes` parameter?

RodolfoSilva (changes requested):

packages/stacked_services/example/lib/app/routes.dart
sebastianbuechler: Sure, done.

RodolfoSilva (approved):
LGTM!

FilledStacks:
Awesome. Thanks

PR #134, opened 2020-08-21
YAML for localisation files

MrTarantula:
I'd prefer to have localization files as YAML, mainly to have the original language as comments. Ex:

```yaml
# Español

HomeView:
  title: Esta es mi casa # This is my Home
  subtitle: Vivo en esta casa # I live in this Home
```

This PR changes from `dart:convert:json.decode()` to `yaml.loadYaml()`. And since JSON is a subset of YAML, This is not a breaking change. The JSON decodes the same as before.

I updated documentation and added tests for the runner for both JSON and YAML. From now on you will need to run `flutter pub run build_runner test` to test this package. However, there is a bug when running tests::

```
Running tests...

Precompiling executable...
Precompiled test:test.
00:00 +0: test\map_utils_test.dart: MapUtilsTest - getStringKeysCodeFromMap - When given an empty map should return generator comment only
00:00 +1: test\localisation_string_key_generator_test.dart: LocalisationStringKeyGeneratorTest - Assets - When given a JSON file, should generate a class called HomeViewStrings
00:00 +2: test\localisation_string_key_generator_test.dart: LocalisationStringKeyGeneratorTest - Assets - When given a JSON file, should generate a class called HomeViewStrings
00:00 +3: test\localisation_string_key_generator_test.dart: LocalisationStringKeyGeneratorTest - Assets - When given a JSON file, should generate a class called HomeViewStrings
00:00 +4: test\localisation_string_key_generator_test.dart: LocalisationStringKeyGeneratorTest - Assets - When given a YAML file, should generate a class called HomeViewStrings
00:00 +5: All tests passed!
```

If I edit any of the MapsUtils tests, they fail as expected, so I believe this is just a display bug in the `build_test` package.

### DIFF
--- a/packages/stacked_localisation/stacked_localisation/README.md
+++ b/packages/stacked_localisation/stacked_localisation/README.md
@@ -8,7 +8,7 @@ The localisation package is very simple. It provides you with a service from whi
 
 ## Setup
 
-To use the localisation service there's a few things that has to be done. We start off by adding the stacked_localisation and the stacked_localisation_generator package.
+To use the localisation service there's a few things that has to be done. We start off by adding the stacked_localisation and the stacked_localisation_generator package (and build_runner if you don't already have it).
 
 ```yaml
 dependencies:
@@ -17,6 +17,7 @@ dependencies:
 
 dev_dependencies:
   ...
+  build_runner:
   stacked_localisation_generator:
 ```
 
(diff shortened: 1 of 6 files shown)


FilledStacks:
Sorry for taking so long. This is pretty dope! i might have jumped the gun on the localisation, i might use an existing package and wrap it behind my service.

PR #144, opened 2020-09-09
Custom Snackbar Fix

ghost:
Fixed customShowSnackBar, which was not allowing null titles or messages, and also prevented a lot of the other settings from working - such as borderRadius. Some settings weren't using the correct snackbarConfig either. Now works like showSnackBar, but will all the other available options.


packages/stacked_services/lib/src/snackbar/snackbar_service.dart
```diff
@@ -122,26 +122,26 @@ class SnackbarService {
...
         style: TextStyle(
-          color: snackbarConfig?.titleColor ??
-              snackbarConfig?.textColor ??
-              Colors.white,
-          fontWeight: FontWeight.w800,
-          fontSize: 16,
-        ),
-      ),
-      messageText: Text(
+            color: snackbarConfig.titleColor,
```
FilledStacks: When text color is set it won't take effect anymore. This now forces the developer to always supply a titleColor and a messageColor which is not how we're using it at the moment. This also removes the default styling of a grey snackbar with white text.

Is there any reason you would remove existing functionality if it causes no problems?

Please make sure when the user does not supply a titleColor the general textColor is used. If the user does not supply a textColor then white is used. The same goes for the messageColor

FilledStacks (changes requested):
I don't understand why the default styling functionality and fallbacks for developers that don't supply title and message color has been removed.

greyatnine:
As of latest commit, have reintroduced default styling functionality and fallbacks for developers that don't supply title and message color that were previously missed.

ghost:
As of latest commit, have reintroduced default styling functionality and fallbacks for developers that don't supply title and message color that were previously missed.

FilledStacks:
Hey, I'm not going to approve this code. I don't know what changed. It looks like you removed stuff then just added it back. 

Please let me know what the exact problem is then i'll fix it. I can't merge in this code given i can't see any value that would be gained by the end user of this package.

ghost:
> Hey, I'm not going to approve this code. I don't know what changed. It looks like you removed stuff then just added it back.
> 
> Please let me know what the exact problem is then i'll fix it. I can't merge in this code given i can't see any value that would be gained by the end user of this package.

I understand. This was my first time using GitHub so I made some mistakes in the process. The only relevant lines changed were between 125 and 144. All other changes were just code format changes made accidentally, which has no impact on the actual code.

In the current Stacked Services, showCustomSnackBar can't be called without a title as it produces a 'A non-null String must be provided to a Text widget.' error. This pull request fixes that.

FilledStacks:
I see. In what scenario will you want to show a snackbar without a title? Maybe i'm missing it.

ghost:
> I see. In what scenario will you want to show a snackbar without a title? Maybe i'm missing it.

If you look at the Material Design page for snackbars you'll see that none of them have a title. Also, when using .showSnackbar, the title is optional there too. 

Limiting what users can build with a custom snackbar doesn't seem correct - especially when it is so inconsequential

carlomigueldy:
is this pull request gonna get merged anytime soon?

FilledStacks:
No. This should actually be closed, the PR did additional things that I don't want to change. I'll go over the thread when I get time and implement what's required to get the functionality the initial PR was meant for.

SudoSharma:
> > I see. In what scenario will you want to show a snackbar without a title? Maybe i'm missing it.
> 
> If you look at the Material Design page for snackbars you'll see that none of them have a title. Also, when using .showSnackbar, the title is optional there too.
> 
> Limiting what users can build with a custom snackbar doesn't seem correct - especially when it is so inconsequential

@FilledStacks thank you for all the hard work you are putting into stacked_services and stacked. These are extremely useful to the community, for beginners and seasoned flutter devs alike. 

I wanted to second what @greyatnine said about making the `title` field optional when using `showCustomSnackBar`. I'd love to have the option to remove the title and have a more "lean" snackbar. Example: https://material.io/components/snackbars

While it's possible to remove the `title` with `showSnackbar`, the borderRadius isn't right, and I'd like to have the flexibility of `showCustomSnackBar`. 

Just wanted to add my support for if/when you get around to reviewing this thread! Thanks again.

FilledStacks:
@SudoSharma I'll merge this in now review and publish. Should be up in about 30 minutes or so.

PR #170, opened 2020-10-12
reapplying system theme when app resumes

abdul10ahad:
Reapply system theme to solve the status bar color issue whenever the app resumes


packages/stacked_themes/lib/src/theme_builder.dart
```diff
@@ -58,4 +60,62 @@ class _ThemeBuilderState extends State<ThemeBuilder> {
...
+    WidgetsBinding.instance.removeObserver(this);
+  }
+
+  @override
+  void didChangeAppLifecycleState(AppLifecycleState state) {
+    super.didChangeAppLifecycleState(state);
+
+    switch (state) {
+      case AppLifecycleState.inactive:
+        print('appLifeCycleState inactive');
```
FilledStacks: Please remove the print statements or make it optional

packages/stacked_themes/lib/src/theme_builder.dart
```diff
@@ -58,4 +60,62 @@ class _ThemeBuilderState extends State<ThemeBuilder> {
...
+  }
+
+  //NOTE: re-apply the appropriate theme when the application gets back into the foreground
+  void adjustSystemThemeIfNecessary() {
+    switch (themeManager.selectedThemeMode) {
+      //do nothing
+      case ThemeMode.dark:
+        break;
+      //do nothing
+      case ThemeMode.dark:
```
FilledStacks, line 107: 2 cases for `ThemeMode.dark` one of them has to be `ThemeMode.light`

packages/stacked_themes/lib/src/theme_builder.dart
```diff
@@ -58,4 +60,62 @@ class _ThemeBuilderState extends State<ThemeBuilder> {
...
+        break;
+      //reapply theme
+      case ThemeMode.system:
+        print('reApplying system theme');
```
FilledStacks: Please remove the log or make it optional.

FilledStacks (changes requested):
Looks good, just the logs that I would prefer to be optional. The way that optional logs are usually implemented is by supplying a verbose bool flag to the object printing it out. The bool should be false be default and the logs should only print if it's true. 

Another thing to do is format the logs properly. To avoid cluttering the user of the packages logs you should change the format to the following. 

NameOfClass - Function Name - Log Message

This will help the user determine where the log is coming from and make it much easier to search for when looking for bugs. In our case the logs would be 

```
print('ThemeManager - didChangeAppLifecycleState - appLifeCycleState resumed')
```

packages/stacked_themes/lib/src/theme_builder.dart
abdul10ahad: thanks for the feedback I'll update the pull request

abdul10ahad:
I've made the changes required

FilledStacks:
Thanks @frodo10messi  will bump the version, test it out and then push if it's ready.

PR #193, opened 2020-12-23
Exposing  color parameters for action buttons (Dialog Service)

elhe26:
This PR allows us to change colors for action buttons on dialog service.


FilledStacks:
This introduces UI code into the VIewMoels by passing in a Color object from the ViewModel or another service which goes against the stacked rules. 

When I get the time I will see if I can devise a way to set the styles dynamically per dialog.

elhe26:
Excellent!

FilledStacks:
@elhe26 I'm tracking this issue [here](https://github.com/FilledStacks/stacked/issues/202). It will be completed in next week. Sorry for the delay.

FilledStacks:
@elhe26 actually I think I'm being a bit too romantic around "No UI code in the viewmodel and services". The color object is simply just a model to describe a color. It doesn't actually require Flutter to run. I'll merge this in now.

PR #205, opened 2021-01-22
👌 IMPROVE: Add isScrollControlled parameter for custom bottom sheets.

Prn-Ice:
This parameter helps with fixing the issue described [here](https://stackoverflow.com/questions/48968176/how-do-you-adjust-the-height-and-borderradius-of-a-bottomsheet-in-flutter), changing the height of the bottom sheet.


FilledStacks:
Thanks @Prn-Ice I'll make a release later today with the new update.

PR #208, opened 2021-01-31
WIP Null-Safety Migration

sameeerkashyap:
Migrating Stacked packages to null safety, 

- [ ] stacked
- [x] stacked_hooks 
- [ ] stacked_services
- [ ] stacked_localisation
- [ ] stacked_themes


FilledStacks:
Awesome, thanks man! When everything is ready we will push it all together.

sameeerkashyap:
@FilledStacks  Need your help, I'm trying to migrate `stacked_themes` but looks like there is a change in `StreamProvieder`, 
it now expects an initialData: argument. but it wasn't there. What Can I pass in?

Screen Shot 2021-02-11 at 9 40 25 AM

FilledStacks:
@Sameerkash mmmmmm,that's a tricky one. Right off the top of my head I'd say you should add a property on the ThemeManager that exposes a themeModel, this would construct a themeModel and return either the values from the shared preferences that's been saved or the default values of the theme. 

Does that make sense, if it doesn't I'm sorry. I'm not looking at the code. just trying to give feedback at the moment based on what I can remember.

sameeerkashyap:
@FilledStacks Yeah I think I got it, 

From the `ThemeManager` return a `ThemeModel` which acts returns cache or default. 
This can be a getter maybe ? 

What do I give as defaults ? 

Although I was questioning the implementation within provider. 
Shouldn't `intialData` be optional?

FilledStacks:
@Sameerkash yes it should be, the user should be allowed to get a null if they want it. And there definitely shouldn't need to be initial data. 

Defaults could be determined by what the user supplies maybe? I think they construct the Thememanager with data, we could use that or just go with `system` for the theme type and then pass the light/dark theme if they supplied it.

sameeerkashyap:
@FilledStacks okay. I'll check it out and get back here by Sunday maybe.

FilledStacks:
@Sameerkash I'll merge in the rest as you make a PR. I just want to check the progress and see if I can help with anything.

FilledStacks:
i don't know what I'm missing here but no packages has been migrated to null-safety yet? When I change the pubspec sdk min to 2.12 then it's all red everywhere 😅 maybe you didn't push the latest changes? @Sameerkash

sameeerkashyap:
> i don't know what I'm missing here but no packages has been migrated to null-safety yet? When I change the pubspec sdk min to 2.12 then it's all red everywhere 😅 maybe you didn't push the latest changes? @Sameerkash 

Yes I didn't push any of the latest changes I did. 

I'll do it soon in a few minutes.

PR #212, opened 2021-02-04
Uses unique instance of GetIt locator to avoid conflicts with apps

josh-burton:
GetIt supports creating individual instances. When using GetIt inside of a package this is good practice, otherwise conflicts can arise between an app using GetIt and the package.


FilledStacks:
Thanks!

FilledStacks:
Pushed in 0.2.3

PR #215, opened 2021-02-08
Made dialog title or content optional

Endagegnehu:
Made dialog title or content optional:

- Handled dialog when title or content is not provided by showing one that is provided.


packages/stacked_services/lib/src/dialog/platform_dialog.dart
```diff
@@ -112,31 +112,75 @@ class PlatformDialog extends StatelessWidget {
...
-            style: TextStyle(color: isDarkMode ? Colors.white : Colors.black),
-          ),
-          actions: actions,
-        );
+        if (content == null) {
+          return AlertDialog(
+            backgroundColor: isDarkMode ? Colors.grey[850] : Colors.white,
+            titleTextStyle: Theme.of(context).dialogTheme.titleTextStyle,
+            contentTextStyle: Theme.of(context).dialogTheme.contentTextStyle,
+            title: Text(
```
FilledStacks: This is a very bad implementation. We should instead just pass null to the named parameter if it's null.

```dart
AlertDialog(
  title: title != null ? Text(title,..) : null,
  content: content != null ? Text(content, ...): null
)
```

packages/stacked_services/lib/src/dialog/platform_dialog.dart
```diff
@@ -112,31 +112,75 @@ class PlatformDialog extends StatelessWidget {
...
+              style: TextStyle(color: isDarkMode ? Colors.white : Colors.black),
+            ),
+            actions: actions,
+          );
+        }
+    }
+  }
+
+  CupertinoAlertDialog buildCupertinoAlertDialog() {
+    if (title == null) {
```
FilledStacks: Apply the same functionality for this code as what was explained above. Having so much duplicated code is never a good sign. It has to be avoided.

FilledStacks (changes requested):
Remove all the duplicated code and replace with better code

FilledStacks (approved):
Much better. Will merge in and release now

PR #225, opened 2021-02-23
Update pubspec.yaml

Ty4Lom:
To allow stacked_generator to work with freezed 0.12.1 and build_runner >=1.10.6


FilledStacks:
Thanks. Pushed in 0.1.3

PR #235, opened 2021-03-01
Fixes issue/171

bats64mgutsi:
Details about the fix can be found in the commit.


FilledStacks (approved):
This works because Consumer automatically disposes the provider at hand. 

But this might break the nonReactive functionality given that the Consumer will rebuild when change notifier is called? i'll have to check that but for now we can approve since it does fix the problem.

PR #237, opened 2021-03-01
Fixes #232

### LINKED ISSUE #232: Incompatible with code-generators that uses different versions of analyser
Hi,
 
 I have faced the next issue:

> Because every version of stacked_generator depends on analyzer ^0.39.2 and floor_generator 0.18.0 depends on analyzer ^0.40.4, stacked_generator is incompatible with floor_generator 0.18.0.
> And because no versions of floor_generator match >0.18.0 <0.19.0, stacked_generator is incompatible with floor_generator ^0.18.0.
> So, because *my_app_name* depends on both floor_generator ^0.18.0 and stacked_generator any, version solving failed.
> pub get failed (1; So, because *my_app_name* depends on both floor_generator ^0.18.0 and stacked_generator any, version solving failed.)

I am trying to use stacked_generator but I am getting the issue above.

To solve it, I have tried to add the next code:

```
dependency_overrides:
  analyzer: ^0.41.2
```

When I try to run:

> flutter pub run build_runner build --delete-conflicting-outputs

I always get the next issue: 

>  Failed assertion: boolean expression must not be null

As you see, I would like to use Floor (https://pub.dev/packages/floor) as ORM.

stacked_generator and Floor use different versions of analyzer so I am not sure but it might be a reason. I think it could be solved by increased the min version on the analyzer library in stacked_generator. Also, the additional reason could be the support of NullSafety

I am new in Flutter, so I could be totally wrong. Please, could you help me? Thanks a lot!

CoryADavis:
I had the similar problem description as the submitter of Issue #232. The problems stems from using an analyzer version that requires the ```withNullability``` argument on ```.getDisplayString()```. This was an [issue](https://github.com/hivedb/hive/commit/24c4c418b15d12d88cc8400aca8638a03a783558) in the hive codebase as well. Their fix handles an edge case I wasn't personally considering when I fixed it originally, so I implemented their fix.

I uncommented ```.vscode/``` in the gitignore, because I needed to override my user-level settings to not alter the code formatting the stacked package has chosen. I didn't see any conflicts with this ignore amendment within the stacked_generator project.

### DIFF
--- a/packages/stacked_generator/.gitignore
+++ b/packages/stacked_generator/.gitignore
@@ -18,7 +18,7 @@
 # The .vscode folder contains launch configuration and tasks you configure in
 # VS Code which you may wish to be included in version control, so this line
 # is commented out by default.
-#.vscode/
+.vscode/
 
 # Flutter/Dart/Pub related
 **/doc/api/
--- a/packages/stacked_generator/lib/src/generators/getit/stacked_locator_generator.dart
+++ b/packages/stacked_generator/lib/src/generators/getit/stacked_locator_generator.dart
@@ -60,13 +60,13 @@ class StackedLocatorGenerator extends GeneratorForAnnotation<StackedApp> {
 
     throwIf(
       classElement == null,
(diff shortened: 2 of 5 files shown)


FilledStacks (approved):
awesome! Thanks. Merging in and releasing now.

PR #245, opened 2021-03-05
Removes the warning from the get package

nikunjdk:
This will remove the following warning
[38;5;244m[GETX] [39;49mWARNING, consider using: "Get.to(() => Page())" instead of "Get.to(Page())".
       Using a widget function instead of a widget fully guarantees that the widget and its controllers will be removed from memory when they are no longer used.


FilledStacks:
Awesome. Thanks!

PR #259, opened 2021-03-10
Support change notifier in ReactiveServiceMixin

lukaszdebowski:
Disclaimer: this is my first ever contribution to open source. Please leave a comment if I did something wrong ;)

Description:
This PR adds support for `ChangeNotifier` in `ReactiveServiceMixin`, i.e our service can just extend the `ChangeNotifier` and does not have to use `observable_ish` values.

Then, instead of having to pass in different RxValues we can just do this:

`listenToReactiveValues([this])`

inside our ReactiveService class (that has to extend ChangeNotifier though)


FilledStacks:
I like it. The ReactiveServiceMixin already has a notifyListeners function call that we're using. We've moved away from Observable Ish. I'll merge this in and probably push a new release later

PR #266, opened 2021-03-11
Replaced flutter_facebook_login with flutter_facebook_auth

merfire:
Replaced [flutter_facebook_login](https://pub.dev/packages/flutter_facebook_login) with [flutter_facebook_auth](https://pub.dev/packages/flutter_facebook_auth) because it is better maintained and it is used in [FlutterFire Facebook social auth documentation](https://firebase.flutter.dev/docs/auth/social#facebook).

### DIFF
--- a/packages/stacked_firebase_auth/lib/src/firebase_authentication_service.dart
+++ b/packages/stacked_firebase_auth/lib/src/firebase_authentication_service.dart
@@ -201,6 +201,8 @@ class FirebaseAuthenticationService {
         case FacebookAuthErrorCode.FAILED:
           return FirebaseAuthenticationResult.error(
               errorMessage: 'Facebook login has failed');
+        default:
+          return FirebaseAuthenticationResult.error(errorMessage: e.toString());
       }
     } catch (e) {
       log?.e(e);
--- a/packages/stacked_firebase_auth/lib/src/firebase_authentication_service.dart
+++ b/packages/stacked_firebase_auth/lib/src/firebase_authentication_service.dart
@@ -191,9 +191,18 @@ class FirebaseAuthenticationService {
       return await _handleAccountExists(e);
     } on FacebookAuthException catch (e) {
       log?.e(e);
-      if(e.errorCode == FacebookAuthErrorCode.CANCELLED) return null;
-
-      return FirebaseAuthenticationResult.error(errorMessage: e.toString());
+      switch (e.errorCode) {
+        case FacebookAuthErrorCode.OPERATION_IN_PROGRESS:
+          return FirebaseAuthenticationResult.error(
+              errorMessage:
+                  'You have a previous Facebook login operation in progress');
+        case FacebookAuthErrorCode.CANCELLED:
+          return FirebaseAuthenticationResult.error(
+              errorMessage: 'Facebook login has been cancelled by the user');
+        case FacebookAuthErrorCode.FAILED:
+          return FirebaseAuthenticationResult.error(
+              errorMessage: 'Facebook login has failed');
+      }
     } catch (e) {
       log?.e(e);
       return FirebaseAuthenticationResult.error(errorMessage: e.toString());
--- a/packages/stacked_firebase_auth/lib/src/firebase_authentication_service.dart
+++ b/packages/stacked_firebase_auth/lib/src/firebase_authentication_service.dart
@@ -73,7 +73,7 @@ class FirebaseAuthenticationService {
         idToken: googleSignInAuthentication.idToken,
       );
 
-      var result = await _signInWithCredential(credential);
+      final result = await _signInWithCredential(credential);
 
       // Link the pending credential with the existing account
       if (_pendingCredential != null) {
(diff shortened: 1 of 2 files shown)


packages/stacked_firebase_auth/lib/src/firebase_authentication_service.dart
```diff
@@ -203,6 +189,11 @@ class FirebaseAuthenticationService {
     } on FirebaseAuthException catch (e) {
       log?.e(e);
       return await _handleAccountExists(e);
+    } on FacebookAuthException catch (e) {
+      log?.e(e);
+      if(e.errorCode == FacebookAuthErrorCode.CANCELLED) return null;
```
FilledStacks: This library will always return a `FirebaseAuthenticationResult`. Please return `FirebaseAuthenticationResult.error(errorMessage: 'Facebook login has been cancelled by the user');` instead of null.

FilledStacks (changes requested):
Definitely way less code so I'm happy with this change. Thanks for making the PR.

packages/stacked_firebase_auth/lib/src/firebase_authentication_service.dart
merfire: If we don't want to return null here I can just remove `if(e.errorCode == FacebookAuthErrorCode.CANCELLED) return null;` and `return FirebaseAuthenticationResult.error(errorMessage: e.toString());` will handle it.

Or we can handle all FacebookAuthExceptions like this:
```dart
      switch (e.errorCode) {
        case FacebookAuthErrorCode.OPERATION_IN_PROGRESS:
          return FirebaseAuthenticationResult.error(
              errorMessage:
                  'You have a previous Facebook login operation in progress');
        case FacebookAuthErrorCode.CANCELLED:
          return FirebaseAuthenticationResult.error(
              errorMessage: 'Facebook login has been cancelled by the user');
        case FacebookAuthErrorCode.FAILED:
          return FirebaseAuthenticationResult.error(
              errorMessage: 'Facebook login has failed');
      }
```

A question about Google and Apple sign in, they both have return null on cancel:

Google:
```dart
      if (googleSignInAccount == null) {
        log?.i('Process is canceled by the user');
        return null;
      }
```
Apple:
```dart
    on SignInWithAppleAuthorizationException catch (e) {
      if (e.code == AuthorizationErrorCode.canceled) return null;

      return FirebaseAuthenticationResult.error(errorMessage: e.toString());
    } 
```

Maybe this should be changed in another PR?

FilledStacks: Good catch. We can probably change that in this PR as well if you don't mind.

merfire:
@FilledStacks
I replaced `return null;` with `FirebaseAuthenticationResult.error(...)` for Google and Apple error handling.

For Google it returns:
`return FirebaseAuthenticationResult.error(errorMessage: 'Google Sign In has been cancelled by the user');`

For Apple I removed `if (e.code == AuthorizationErrorCode.canceled) return null;`
because the line below `return FirebaseAuthenticationResult.error(errorMessage: e.toString());`
will handle all errors for Apple Sign In.

`e.toString()` for Apple errors returns string in this format:
'SignInWithAppleAuthorizationError($code, $message)'

Is this ok or should we just return message here without the code?

FilledStacks (approved):
Thank you. Will merge when tested in a project.

FilledStacks:
@merfire does this package any requirements to the code base using it? Meaning if you include the package in a project that doesn't have FB setup in the Android manifest does it still work?

I don't want to force users to have to use Facebook auth.

merfire:
@FilledStacks 
I just tested this. It won't fail until `signInWithFacebook()` function is called, which will return:
`'MissingPluginException(No implementation found for method login on channel app.meedu/flutter_facebook_auth)'`
if there is no FB setup.

One thing I realized in this test for some reason I got this error:
```
Execution failed for task ':app:checkDebugDuplicateClasses'.
> A failure occurred while executing com.android.build.gradle.internal.tasks.CheckDuplicatesRunnable
   > Duplicate class com.google.common.util.concurrent.ListenableFuture found in modules jetified-guava-20.0 (com.google.guava:guava:20.0) and jetified-listenablefuture-1.0 (com.google.guava:listenablefuture:1.0)
```
which can be resolved by adding this:
`implementation 'com.google.guava:listenablefuture:9999.0-empty-to-avoid-conflict-with-guava'`
to dependencies in android/app/build.gradle.

Example:
```
dependencies {
    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
    implementation 'com.android.support:multidex:1.0.3'
    implementation 'com.google.guava:listenablefuture:9999.0-empty-to-avoid-conflict-with-guava'
}
```

_Note: multidex was added here because it is in the documentation for Firebase Core._

FilledStacks:
Perfect @merfire I will pull this down and push an update for it today.

FilledStacks:
Pushed in 0.1.3

Issue #267, opened 2021-03-11
[stacked_firebase_auth] FirebaseAuthenticationResult has only uid and userToken information


merfire:
All sign-in methods (email, google, apple, facebook) return `FirebaseAuthenticationResult` object that has only uid, userToken and errorMessage.
Is there a reason why only uid and userToken are returned and not the whole User object from FirebaseAuth that has uid, userToken, email, displayName, phoneNumber, photoUrl, and most importantly customClaims?
Is it better to return User object in FirebaseAuthenticationResult ?
And is it better to expose _firebaseAuth, that way all functions of FirebaseAuth.instance can be accessed directly?

mrseth01:
I usually listen for `authService.authChanges.listen((user) {}` to get the user object and sync custom user profiles stored in Firestore.

FilledStacks:
@merfire No reason the user object is not on there. I usually return only what's needed at the time and add as those requirements change. This way I know exactly what's needed. I don't want code exposed that won't be used. 

It could be better to return the user object and have properties on there for quick access. 

We could expose firebase auth as well. If you'd like you can make a PR I can review, otherwise I'll most likely only get to this in a week or two from now.

merfire:
I can do this after finishing Facebook changes for stacked_firebase_auth.

FilledStacks:
@merfire that would be great thanks.

FilledStacks:
we now return the user in the Firebase Auth Result

PR #271, opened 2021-03-22
[stacked_firebase_auth] Return firebase user, exposed firebaseAuth for easier access, bumped versions for Google, Apple and Facebook plugins

merfire:
Related to #267 discussion

Changed FirebaseAuthenticationResult to return Firebase User object instead of only uid and userToken.

Made firebaseAuth public for easier access from FirebaseAuthenticationService.

Bumped versions to latest for Google, Apple and Facebook plugins. 
This was needed for fixing a version solving failed problem with the latest firebase_core and current flutter_facebook_auth.
```
Because flutter_facebook_auth >=2.0.0-web.8 <3.0.0-nullsafety.0 depends on flutter_facebook_auth_platform_interface ^1.0.1 which depends on plugin_platform_interface ^1.0.1, flutter_facebook_auth >=2.0.0-web.8 <3.0.0-nullsafety.0 requires plugin_platform_interf0-nullsafety.0 requires plugin_platform_interface ^1.0.1.
And because firebase_core >=1.0.0 depends on firebase_core_platform_interface ^4.0.0 which depends on plugin_platform_interface ^2.0.0, flutter_facebook_auth >=2.0.0-web.8 <3.0.0-nullsafety.0 is incompatible with firebase_core >=1.0.0.rebase_core >=1.0.0.
So, because stacked_firebase_auth depends on both firebase_core ^1.0.1 and flutter_facebook_auth ^2.0.2+1, version solving failed.
```

### DIFF
--- a/packages/stacked_firebase_auth/lib/src/firebase_authentication_service.dart
+++ b/packages/stacked_firebase_auth/lib/src/firebase_authentication_service.dart
@@ -27,7 +27,7 @@ class FirebaseAuthenticationService {
   /// Usually a reverse domain notation like com.example.app.service
   final appleClientId;
 
-  final _firebaseAuth = FirebaseAuth.instance;
+  final firebaseAuth = FirebaseAuth.instance;
   final _googleSignIn = GoogleSignIn();
   final _facebookLogin = FacebookAuth.instance;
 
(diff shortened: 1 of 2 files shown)


merfire:
Maybe we can add
`export 'package:firebase_core/firebase_core.dart';`
to stacked_firebase_auth.dart for easier access to `Firebase.initializeApp()`?

This will enable autocomplete for Firebase Core everywhere stacked_firebase_auth is imported and there will be no need to manually import firebase_core.

PR #277, opened 2021-03-25
Adds date form field to stacked forms (Null Safe)

bats64mgutsi:
At the moment we can use a basic form using AppSkeletons that takes in text input and passes that to the viewmodel for submission. This should be updated to take in more than just the forms we have now. This will be done using the Forms functionality in stacked, and stacked might have to be updated to support more than just text as well.

- [x] Date Selection
- [ ] Static dropdown list selection
- [ ] large text form field


FilledStacks (approved):
This looks good! Can't wait to try it out. I'll publish a new null-safety version now.

PR #302, opened 2021-04-02
Removed `flutter_facebook_auth` dependency and its implementation

carlomigueldy:


FilledStacks (changes requested):
The version isn't bumped in the `pubspec.yaml` file please change it to match the one in the Changelog

carlomigueldy:
I've adjusted the version under `pubspec.yaml` and from the change log as well :)

FilledStacks (approved):
This had to be done. It was a terrible bug that was happening. We can add it back when the package allows for it all to be used together.

merfire:
May I ask what happened here? Was this failing on iOS builds?
In my tests on Android, `MissingPluginException` only occurred when I explicitly called `signInWithFacebook()` function without configuration.

FilledStacks:
@merfire this was removed because when you try to call google auth login you get a "not implemented exception" if facebook auth is not setup.

PR #305, opened 2021-04-02
Changed ViewModel variable name to be consistent #301

### LINKED ISSUE #301: The ViewModel variable name is inconsistent
The ViewModel variable name is inconsistent in the stacked package. 

`ViewModelBuilder` uses just `model`. For example:

```dart
return ViewModelBuilder<HomeViewModel>.reactive(
    builder: (context, model, child) => Scaffold() 
);
```

While the `ViewModelWidget` uses `viewModel`. For example: 

```dart
class Widget extends ViewModelWidget<HomeViewModel> {
  @override
  Widget build(BuildContext context, HomeViewModel viewModel) {
    return Container();
  }
}
```

@FilledStacks I can create a PR to make it consistent, would you have a preference for which variable name to use? I think I prefer just `model` and it seems like you do as well looking at the readme.

Moseco:
I went through and updated the variable naming scheme to match `ViewModel viewModel`.

Also updated the documentation to reflect that change.

Closes #301

### DIFF
--- a/packages/stacked/CHANGELOG.md
+++ b/packages/stacked/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.0.2
+
+- Made ViewModel variable naming scheme consistent. `ViewModel viewmodel` is now preferred.
+
 ## 2.0.1
 
 - Adds ReactiveValue and ReactiveList for reactive functionality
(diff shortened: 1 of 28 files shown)


FilledStacks (approved):
This should have been done a long time ago. Thank you.

Issue #198, opened 2021-01-16
[Stacked-Services] Barrier dismissible not working for bottomSheetService


akifarhan:
I tried to use `showCustomSheet` and set the `barrierDismissible` to `false` but it is not working. I've checked the source code and found that the `barrierDismissible` was not assigned to Get's `isDismissible`. I'm not sure either it is intentionally or not. 

![image](https://user-images.githubusercontent.com/38977468/104818463-cfb95100-5862-11eb-951c-00109f575731.png)

I wanted to create a PR but it seems like Stacked-Services gonna get a major changes so I'm kinda afraid. 😅

FilledStacks:
haha, no man. Please make the PR from the latest version of the code and I'll merge that in. The way I make the services is to only expose what's needed. When a dev, like yourself, comes a long and needs something we add that in :) . I'll tag this issue with Waiting for PR if you're happy with that?

FilledStacks:
We're going to tackle this and add it into the code.

akifarhan:
I think this issue has been solved @FilledStacks through [this PR](https://github.com/FilledStacks/stacked/pull/203). I forgot to close this issue

PR #307, opened 2021-04-03
This fixes issue #198

hurbes:
Fixed barrierDismissible issue on BottomSheetService
Added exitBottomSheetDuration and enterBottomSheetDuration


FilledStacks (changes requested):
Great! 

Just one change. Update the version to 0.8.2 instead. Latest is on 0.8.1, I pushed earlier. 

Pull that into your branch and then add the 0.8.2 above that in the change log.

hurbes:
done

FilledStacks (approved):

Issue #218, opened 2021-02-14
[Stacked Services] currentRoute is always empty


bdairy:
Hello there,

I am using the navigation service in my app..

I have a view (QR scanner view) that is accessable  from 2 places.. 

the first place is as a startup view,,, in this case I am showing a button to take me to the dashboard..

the second place is in the dashboard...  in this case I am not showing a button to take me to the dashboard, I am just using the back button.

when I access the view from the dashboard simply by calling 
`locator<NavigationService>().navigateTo(QRScannerRoute);`

In the QRScannerView  I want to check if my currentRoute is empty _**"which means It is the startup view"**_ Show the link to take me to the dashboard... but if my current route is not empty _**which indicates I accessed it from the dashboard**_  just use the default backButton... But in all the cases currentRoute is always empty,,, same applies to the previous route..

To summarise... 

`locator<NavigationService>().navigateTo(QRScannerRoute);`

leads to empty currentRoute in the target view always.

also I couldn't find a way to access the navigation stack to check if it is empty or not..

FilledStacks:
@bdairy we use get's current route functionality directly so we don't have any code around how that works in the package. We index into `Get.currentRoute` . Can you put your main.dart file in here. I'd like to check for setup errors.

bdairy:
@FilledStacks  I admire how you care about your product, you deserve all the support..

Main.dart file is quite big and chunked into many files,,  I am using a scoped model wrapper for the whole app for localization purposes, and theme is separated..
The nav service is working  except the mentioned issue.
I'll try to put what might be relevance to the navigation, I hope it helps:

**main.dart**

```
void main() {
  setupLocator();
  setupDialogUi();
  setupBottomSheets();

  runApp(new ScopeModelWrapper());
}
```


**My App**

```
class MyApp extends StatelessWidget {
  // This widget is the root of your application. Main App separated with scoped model descendent
  @override
  Widget build(BuildContext context) {
    return FutureBuilder(
        future: Firebase.initializeApp(),
        builder: (context, snapshot) {
          if (snapshot.hasError) {
            return Center(child: Text('Invalid App'));
          }

          // Once complete, show your application
          if (snapshot.connectionState == ConnectionState.done) {
            FlutterError.onError =
                FirebaseCrashlytics.instance.recordFlutterError;
            if (kDebugMode) {
              // Force disable Crashlytics collection while doing every day development.
              // Temporarily toggle this to true if you want to test crash reporting in your app.
[...]

PR #320, opened 2021-04-09
This fixes issue #218

hurbes:
Fixed where currentRoute in NavigationService was always null


FilledStacks (changes requested):
Awesome!

Can you add a section in the readme to describe this. Add under the navigation service section

### Route observation

If you want the current route to be set during navigations then you have to add a route observer. Then put the sample code there that you added into the example file. 

This is great!

hurbes:
updated

FilledStacks (approved):

PR #340, opened 2021-04-21
Adds `emailExists` method

carlomigueldy:
To check if provided email has a corresponding user registered


FilledStacks (changes requested):
Please update the package patch number and add a change into the change log.

FilledStacks (approved):

PR #351, opened 2021-05-07
Stacked Crashlytics

Eimen2018:


packages/stacked_crashlytics/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+## [0.0.1] - TODO: Add release date.
```
FilledStacks: Change this to 0.1.0 and remove the text on the right

For the Entry you can add

- Initial implementation of Crashlytics Service

packages/stacked_crashlytics/LICENSE
```diff
@@ -0,0 +1 @@
+TODO: Add your license here.
```
FilledStacks: Go to the stacked package, copy the licence, make it 2021 for `FilledStacks` and not Dane Mackier

packages/stacked_crashlytics/README.md
```diff
@@ -0,0 +1,14 @@
+# stacked_crashlytics
+
+A new Flutter package project.
```
FilledStacks: Remove the basic flutter readme. Instead add some basic details on how to use. 

Make the title 

#Stacked Crashlytics Service

A service accompanied with a LogOutput that tracks all warnings as non-fatal crash reports and all errors as fatal crash reports. This makes use of the Firebase Crashlytics service.

## How To use

- Go over adding the package
- Go over adding the service to the locator. 
- Show where to add the log output

And that should be good enough.

packages/stacked_crashlytics/pubspec.yaml
```diff
@@ -0,0 +1,57 @@
+name: stacked_crashlytics
+description: A service class that provides Firebase Crashlytics Functionality on a single api.
+version: 0.0.1
```
FilledStacks: Change to 0.1.0

packages/stacked_crashlytics/pubspec.yaml
```diff
@@ -0,0 +1,57 @@
+name: stacked_crashlytics
+description: A service class that provides Firebase Crashlytics Functionality on a single api.
```
FilledStacks: A package that provides a Crashlytics Service as well as a LogOuput

FilledStacks (changes requested):
Some basic package maintenance changes required

packages/stacked_crashlytics/README.md
```diff
@@ -1,14 +1,36 @@
...
+    @lazySingleton
+  CrashlyticsService get crashlyticsService;
+    ```
+    * If you are presolving into an async instance, you can add it inside the file like so
+    ```dart 
+    @preResolve
+  Future<CrashlyticsService> get crashlyticsService =>
+      CrashlyticsService.getInstance();
+    ```
+    Note: Don't forget to run
```
FilledStacks: We should be using the `StackedApp` annotation for this.

FilledStacks (changes requested):
We should use Stacked app for locator setup, not injectable.

FilledStacks (approved):

Issue #350, opened 2021-05-06
[stacked_service] showSnackbar has empty space if title is not given 


alanchan-dev:
stacked_service: 0.8.5

If I don't give the title, there is an empty space at the snackbar, which isn't there on older version.
Can the empty space be removed when there is no title passed?

`_snackbarService.showSnackbar(
        message:
            'message content here');`

Thank you.

FilledStacks:
@DogeeeXD It can probably be removed. I'll see if we have a dev free to tackle that.

FilledStacks:
Fixed in 0.8.6 will push release later today

alanchan-dev:
Thank you very much!

PR #358, opened 2021-05-11
Issue-350

hurbes:
- This fixes issue #350
- The `title` is now a required parameter on `showSnackbar` to avoid the empty space


FilledStacks:
@hurbes The title should not be required. It use to be and then we got requests to make it optional. Is there any way we can collapse that space if it's empty?

hurbes:
The `Get.snackbar` takes positional parameter as a string so I thought the `title` might a required field. I changed back to a default empty string and used a `SizedBox` instead of null.

Screenshot_1620801634

PR #360, opened 2021-05-11
Issue 347

hurbes:


packages/stacked_generator/lib/src/generators/getit/stacked_locator_content_generator.dart
```diff
@@ -22,8 +22,9 @@ class StackedLocatorContentGenerator extends BaseGenerator {
         .any((service) => service.type == DependencyType.PresolvedSingleton);
 
     writeLine(
-        '${hasPresolve ? 'Future' : 'void'} setupLocator() ${hasPresolve ? 'async' : ''} {');
-
+        '${hasPresolve ? 'Future' : 'void'} setupLocator({String? environment , EnvironmentFilter? environmentFilter}) ${hasPresolve ? 'async' : ''} {');
+    writeLine(
+        'final stacked = StackedLocator(environment: environment, environmentFilter: environmentFilter);');
```
FilledStacks: Why has the name changed to stacked here? 

Most apps use the term locator. This is going to cause a breaking change.

FilledStacks (changes requested):
Double checking that the locator name change is not going to be a breaking change for apps that use the stacked_generator

packages/stacked_generator/lib/src/generators/getit/stacked_locator_content_generator.dart
hurbes: It's just the local variable within the `setupLocator` to register the dependencies with, the global variable stays the same as `locator`.

```dart
  writeLine('final locator = StackedLocator.instance;');
 ```

hurbes:
I changed the implementation a bit and moved all the `Environment` related stuff from the constructor to its own function.

FilledStacks (approved):

PR #374, opened 2021-05-17
Use the default filter for the logger

Hannnes1:
This is instead of using LogAllTheTimeFilter which always logged all levels and ignored the level variable in the Logger class.

### DIFF
--- a/packages/stacked_generator/CHANGELOG.md
+++ b/packages/stacked_generator/CHANGELOG.md
@@ -1,5 +1,9 @@
 # ChangeLog
 
+## 0.4.7
+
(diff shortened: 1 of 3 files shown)


FilledStacks (approved):

FilledStacks:
@Hannnes1 Please bump the patch version and add an entry into the Changelog file.

Hannnes1:
I realized that I'm not sure if you got a notification that I have made a new commit. Anyway, I have bumped the version and updated the changelog.

FilledStacks (approved):

PR #406, opened 2021-06-09
Expose FirebaseAuth's authStateChange 

bejaeger:
To mock out authStateChanges() for unit testing, it would be useful to expose it through the stacked FirebaseAuthenticationService. Please, Let me know if that makes sense. Thanks!


FilledStacks:
That does make sense

FilledStacks:
Will release later today

PR #407, opened 2021-06-10
fix #397

### LINKED ISSUE #397: Can't build a dropdown list dynamically because the constructor for StaticDropdownItem is constant
For example we may want to build a dropdown list from a list of countries. 
```
const countries = <String>["Mordor", "Wakanda"];

final countriesDropdownlist = countries
    .map((countryName) =>
        StaticDropdownItem(title: countryName, value: countryName))
    .toList();

@FormView(fields: [
  FormDropdownField(name: 'country', items: countriesDropdownlist),
])
```

If we try to do this we get the following warnings.

Arguments of a constant creation must be constant expressions.
Try making the argument a valid constant, or use 'new' to call the constructor.dartconst_with_non_constant_argument
The values in a const list literal must be constants.
Try removing the keyword 'const' from the list literal.dartnon_constant_list_element
Invalid constant value.dart(invalid_constant)

Please remove the const keyword from the StaticDropdownItem class constructor

kayzkenny:
remove the const keyword from the class constructor


FilledStacks:
This doesn't fix the issue. It actually just makes it so that you can't supply the value to the annoation because it's not const

kayzkenny:
Thanks for the feedback

Issue #404, opened 2021-06-07
stacked_themes build for non-mobile platforms w/ statusBarColorBuilder throws error


calvintam236:
Currently, if you set `statusBarColorBuilder` in the `ThemeBuilder` and try to build for non-Android and non-iOS platform, you will get this error:

```
[ERROR:flutter/lib/ui/ui_dart_state.cc(186)] Unhandled Exception: MissingPluginException(No implementation found for method setstatusbarcolor on channel plugins.sameer.com/statusbar)
#0 MethodChannel._invokeMethod
package:flutter/…/services/platform_channel.dart:156
<asynchronous suspension>
#1 StatusBarService.updateStatusBarColor
package:stacked_themes/…/services/statusbar_service.dart:9
<asynchronous suspension>
#2 ThemeManager._applyStatusBarColor
package:stacked_themes/src/theme_manager.dart:161
<asynchronous suspension>
```

This should not throw error if the target platform does not support it.

FilledStacks:
@hurbes you can implement for the platforms and simply do nothing to avoid the exception being thrown.

FilledStacks:
fixed and merged. Will be released later today.

FilledStacks:
Released in 0.3.4

PR #414, opened 2021-06-15
Fixed Expception being thrown for non-Android and non-iOS platform

hurbes:
This fixes issues #404

### DIFF
--- a/packages/stacked_themes/CHANGELOG.md
+++ b/packages/stacked_themes/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.3.4
+
+- Fixed Expception being thrown for non-Android and non-iOS platform
+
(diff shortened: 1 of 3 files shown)


FilledStacks (approved):

calvintam236:
https://github.com/hurbes/stacked/blob/d2311ba317842e1bd3bace396b98599cb5919f0c/packages/stacked_themes/lib/src/theme_manager.dart#L160

I think it should be `||` not `&&`.

hurbes:
I apologize you're correct, it should be `||` not `&&`. Though it has been already updated, it has been a bad miss on my part.

calvintam236:
@hurbes A hotfix is needed since last change broke everyone’s status bar.

PR #425, opened 2021-06-18
Fix missing items in the list of packages

calvintam236:


FilledStacks (approved):
Much appreciated calvin

PR #430, opened 2021-06-25
Stacked Themes : Bug Fixes | New Features

zainulhassan815:
1- Fixed Platform Checking error when app is run on web as platform isn't supported on web.
2- Added navigation bar color changer
3- Fixed overlay colors ( status bar and navigation bar colors )  when app is resumed


FilledStacks:
Nice PR. I like the feature. 

Please can you run flutter test and post the results below so I can see if all the tests are still passing. 

Also can you bump the patch version and then add the changes into the change log.

zainulhassan815:
Actually I am not very expert but I will surely try. Thanks for such motivating comment.

FilledStacks:
This has gone stale. @zainulhassan815 should I just close it?

Issue #426, opened 2021-06-18
Replace `dart:io` for `universal_io` for better cross-platform support


calvintam236:
I notice some of the packages are still using `dart:io`. Since `stacked` will support web (on the roadmap), this should be replaced by [`universal_io`](https://pub.dev/packages/universal_io) for its cross-platform support (web included) since `dart:io` has no support on web.

`universal_io` has identical API as `dart:io` so the conversion should be quick and easy.

FilledStacks:
@hurbes can you run through the packages and update this for us?

FilledStacks:
released in latest versions

PR #437, opened 2021-07-01
This fixes issues #426

hurbes:
- This fixes issues #426
- Replaced `dart:io` with `universal_io` in `stacked`
- Replaced `dart:io` with `universal_io` in `stacked_themes`


FilledStacks (approved):
Awesome!

PR #444, opened 2021-07-06
Factory with parameter

hurbes:
- Added ability to pass parameters to factories


FilledStacks:
Awesome. @hurbes can you please add into the readme for stacked a section for registering factory params with the locator so that I can see how that will look in code. 

As well as how you use it and pass those params to the factory.

hurbes:
@FilledStacks Done. Updated Readme.

FilledStacks (approved):

PR #447, opened 2021-07-07
Passing down generic types and other properties relevant to get navigation

carlomigueldy:


FilledStacks:
@carlomigueldy thank you for the PR. Just need some package maintenance before I can release. 

Bump the version of stacked and stacked services and then please add to the change log this change as well as the change in the previous PR for adding ViewModel into onDispose callback.

carlomigueldy:
Ok done! @FilledStacks

packages/stacked_services/lib/src/navigation_service.dart
```diff
@@ -96,45 +96,80 @@ class NavigationService {
...
   /// - leftToRightWithFade
   /// - cupertino
-  Future<dynamic>? navigateWithTransition(Widget page,
-      {bool? opaque,
-      String transition = '',
-      Duration? duration,
-      bool? popGesture,
-      int? id}) {
-    return Get.to(
+  Future<T?>? navigateWithTransition<T>(
```
FilledStacks, line 99: If you make this `T` then you can pass in a nullable type or non-nullable type. If you make it `T?` it always has to be a nullable type. We should change it to `T`

FilledStacks (changes requested):
If possible I'd like the type arguments to be changed to T since int? and int are both types.

packages/stacked_services/lib/src/navigation_service.dart
carlomigueldy: @FilledStacks  It's the return type for `T` but not the actual argument, the actual generic type argument is here `navigateWithTransition<T>` so it doesn't have to be always of nullable type

FilledStacks (approved):

PR #452, opened 2021-07-14
[stacked_services] Pass down ignoreSafeArea to Get.bottomSheet

bbedward:
We ran into the issue https://github.com/flutter/flutter/issues/59204 where bottom sheets with `isScrollControlled: true` do not respect SafeArea.

getx has a solution https://github.com/jonataslaw/getx/issues/44

So this change just allows us to pass down `ignoreSafeArea` to the Get.bottomSheet method, which solves our problem with SafeArea.

I'm not sure why you wouldn't want `ignoreSafeArea` to be false all of the time, but either way this does not change existing behavior in case there are implications.

### DIFF
--- a/packages/stacked_services/CHANGELOG.md
+++ b/packages/stacked_services/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.8.11
+
+- Add `ignoreSafeArea` parameter to `showBottomSheet` and `showCustomSheet`, pass down to `get` bottomSheet method
+
 ## 0.8.10
 
 - Pass down generic types to `NavigationService` methods along with the properties relevant to `get` router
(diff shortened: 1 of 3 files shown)


yekta:
To add to this: Not ignoring the safe area causes to not have access to things like `MediaQuery.of(context).padding.top` on the child. If you don't ignore it, `MediaQuery.of(context).padding.top` is 0 on the child. This is problematic because child is the actual sheet. The parent takes away the access to `padding.top` while not actually applying the padding to the sheet. It shouldn't apply the padding to the actual sheet most of the time anyways (if you don't want a floating sheet), it should just let the sheet itself decide.

FilledStacks:
Thank you @bbedward I will merge and push a release today.

FilledStacks (approved):

Issue #455, opened 2021-07-15
[stacked_generator] [logger] `"Error: Bad state: No element"`


saileshbro:
The following exception is being thrown from the logger. Is it because of this? The screenshot of the exception is attached.
https://github.com/flutter/flutter/issues/47040#issuecomment-579185623
![image](https://user-images.githubusercontent.com/43571990/125816036-33d1532b-f2e6-4403-b539-7ae6a68d3e3a.png)

PR #456, opened 2021-07-15
fix: add `orElse` handler in the logger `realFirstLine` var #455

saileshbro:


FilledStacks (changes requested):
Please bump the patch number of the version and add the change into the change log

saileshbro:
#466

PR #466, opened 2021-07-20
dispose focusnodes while calling `disposeForm`

saileshbro:


packages/stacked_generator/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## 0.5.1
 
+- Adds `orElse` handler in the logger `realFirstLine` variable to fix [issue #455](https://github.com/FilledStacks/stacked/issues/455)
```
FilledStacks, line 5: Please move this to a new version in the change log and update the pubspec.yaml to match

FilledStacks (changes requested):
needs a patch version update and change log to reflect under correct version

packages/stacked_generator/CHANGELOG.md
saileshbro: Updated 👍🏻

FilledStacks (approved):

PR #467, opened 2021-07-20
get.generalDialog parameter useRootNavigator

Silfalion:
The get package removed the argument useRootNavigator from the get.generalDialog. This causes the following error:

```
./../snap/flutter/common/flutter/.pub-cache/hosted/pub.dartlang.org/stacked_services-0.8.11/
lib/src/dialog/dialog_service.dart:204:7: Error: No named parameter with the name
'useRootNavigator'.
      useRootNavigator: true,
```

Didn't double check the consequences of deleting so that should be done, sorry about that.


skbhati199 (commented):
Working fine 

useRootNavigator: true, remove this code

FilledStacks:
Removed already in a different PR.

PR #491, opened 2021-08-12
Update stacked firebase auth fire base auth result

EshetuLucas:
Adds FirebaseAuthException Error code to the FirebaseAuthenticationResult.error result.


FilledStacks (changes requested):
Please bump the patch version of the package and add the change into the changelog.

FilledStacks (approved):

PR #493, opened 2021-08-16
Split BaseViewModel into mixin and concrete implementation

safarmer:
This allows more flexibility using the view model in existing class hierarchies.

Also added a method to see if any keys have an error and simplified some code paths:
 - simplified busy check in `skeletonData`
 - return rather than store and return in `runBusyFuture`


FilledStacks:
Hey @safarmer what's the point of this update? Is there any reason it's required.

safarmer:
This is so the view model base class can be used in an existing class hierarchy. There is also a minor simplification of the boolean check in the skeleton data method that makes it much clearer what the expected conditions are (imho).

FilledStacks:
This PR doesn't fall in line with what Stacked wants to offer as a developer experience.

PR #533, opened 2021-11-20
Dependency upgrades

ghost:
Upgraded Provider in stacked and stacked_hooks and google_sign_in in stacked_firebase_auth


FilledStacks (approved):
Thanks @Schefferbird 🎉

PR #542, opened 2021-12-10
:bug: Omitted logs in release mode

rexthecoder:
This PR makes sure logs are omitted in release mode. Based on the logger's recommendation

Capture
 
 I did a reimplementation of the filter class


FilledStacks:
This has been fixed in the latest package

PR #556, opened 2022-01-20
Allows an external NavigatorKey in the DialogService

bw-flagship:


bw-flagship:
It was a bad idea, since it makes more sense to directly configure it via GetX

PR #565, opened 2022-02-07
Fix the generated code of FactoryWithParam with asType

unknown-yuser:
I faced an error of  generated code when I used `FactoryWithParam` with `asType`.
The original generated code will produce like this

```
exampleLocator.registerFactoryParam<FactoryService, String?, double?><IFactoryService>((param1, param2) => FactoryService(key: param1, value: param2));
``` 

I tried to fix and then the revised generated code will produce like this

```
exampleLocator.registerFactoryParam<IFactoryService, String?, double?>((param1, param2) => FactoryService(key: param1, value: param2));
```

### DIFF
--- a/packages/stacked_generator/CHANGELOG.md
+++ b/packages/stacked_generator/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.6.3
+
+- Fixed code generation bug when using `FactoryWithParam` with the `asType` argument set
+
 ## 0.6.2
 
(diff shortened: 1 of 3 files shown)


FilledStacks (changes requested):
Awesome. 

Could you please increase the patch version and add an entry into the changelog that captures this change

FilledStacks:
This functionality has been added to the newest PR that adds testing around all generators. It will be in production soon.

PR #571, opened 2022-02-18
fix: Minor fixes for the snackbar

iamelio:


FilledStacks (changes requested):
I like it. Please can you increase the minor version of the package and add the changes to the changelog

Issue #584, opened 2022-04-14
Add instanceName to get_it generator


zuboje:
The current stacked generator does now allow multiple instances of the same class for a singleton. get_it package has optional parameter "instanceName" that can be passed for multiple singleton instances of the same class:

```
locator.registerLazySingleton(() => MyClass(), instanceName: 'mysetup1');
locator.registerLazySingleton(() => MyClass(), instanceName: 'mysetup2');
```

This allows for greater flexibility. It would be great if stacked annotation would allow an optional parameter to pass the instance name.

FilledStacks:
You think you can add this through a PR?

zuboje:
Certainly I can

zuboje:
I just created PR https://github.com/FilledStacks/stacked/pull/586 but I have no ability to test because I never wrote plugins and I tried using the path in pubspec.yaml and it generates issue for me because I have to changed stacked package for annotation and stacked_generator package. If you give me some quick tip how I can test both at the same time locally I will be more than happy to debug/troublshoot, current code is more like a guess

calvintam236:
I just encountered the same issue. I want to reuse the same service with a different `getInstance` (the library has multiple constructors and method calls do not change).

@FilledStacks It looks like #586 did not merge, and I don't see any code/docs changes regarding this in `master`?

FilledStacks:
This is in the latest generator.

calvintam236:
> This is in the latest generator.

I found the code now, but the readme does not mention the usage of this.

FilledStacks:
Yes, the readme has not been kept up to date, I'll spend some time in the future updating all the docs for Stacked.

PR #586, opened 2022-04-15
Added ability to set instance name in locator

zuboje:
This should allow setting up instanceName in the locator for moredetails take a look at [()](https://pub.dev/packages/get_it#experts-region)

### DIFF
--- a/packages/stacked/lib/src/code_generation/stacked_locator_annotations.dart
+++ b/packages/stacked/lib/src/code_generation/stacked_locator_annotations.dart
@@ -28,6 +28,7 @@ class DependencyRegistration {
   final Function? dispose;
   final Type? param1;
   final Type? param2;
+  final String? instanceName;
 
   const DependencyRegistration({
     this.environments,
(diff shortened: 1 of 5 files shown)


FilledStacks (changes requested):
Can I please see the generated output. I don't think this will compile when generated.

zuboje:
@FilledStacks I will need your guidance to finish it. My app is not allowing me to use two plugins that depend one on another and they are both local it is requiring me for at least 1 plugin to be hosted, otherwise, it does not compile it, just throws an error during the build. How can I simultaneously test "Stacked" plugin changes and "stacked_generator" changes? Stacked changes are required because of the annotation, the generator uses the annotation. Can you please give me some guidance/link to some tutorial on how to test two plugins dependent one on another (generator depends on the stacked plugin)

FilledStacks:
This has been added to the generator with the newest updates to add unit tests.

eb-so:
Closes #584

PR #588, opened 2022-04-21
Update Examples Configs 

Eimen2018:
Fixes 
- Android Embedding to V2 on Example
- Tests that require firebase mock
- updates package versions


FilledStacks (approved):
Awesome! And I see that melos is also added.

Issue #504, opened 2021-09-18
[stacked]Add Object for the key in multiple stream viewmodel


Afsal102:
Adding a specific type of object to the onStreamData => [key] 
To integrate it with unions...
**Sample**
```

onStreamData<Union>(Union key, dynamic data){
	key.when(
			key1:(){
				//do some tasks
		},
	);
}


```
sample is used with **freezed** unions.
(T extends Object) any Type of object

FilledStacks:
Are you asking how to do this or are you asking for a new feature?

Afsal102:
A new feature. I stumbled upon this and found this solution.
Its works great with `unions` also eliminates `if` conditions.
I could make a PR if u would like

Eimen2018:
Hey @Afsal102  Yeah sure you can make a PR we will review it.

Afsal102:
@Eimen2018 sure! am quite busy i will try to make a pr by start of `May` 😄

Afsal102:
@Eimen2018 created a PR 😄

Eimen2018:
@Afsal102  Thanks for the PR

PR #589, opened 2022-04-24
feat: adds custom object for keys in multiple stream view model

Afsal102:
issue: #504

### DIFF
--- a/packages/stacked/CHANGELOG.md
+++ b/packages/stacked/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 2.3.4+0
+- Updates MultipleStreamViewModel to take generic type for key
+
 ## 2.3.3+1
 
 - Updates Example's kotlin version to 1.4.10
--- a/packages/stacked/lib/src/state_management/base_view_models.dart
+++ b/packages/stacked/lib/src/state_management/base_view_models.dart
@@ -230,11 +230,12 @@ class _SingleDataSourceViewModel<T> extends DynamicSourceViewModel {
   bool get dataReady => _data != null && !hasError;
 }
 
-class _MultiDataSourceViewModel extends DynamicSourceViewModel {
-  Map<String, dynamic>? _dataMap;
-  Map<String, dynamic>? get dataMap => _dataMap;
+class _MultiDataSourceViewModel<K extends Object>
+    extends DynamicSourceViewModel {
+  Map<K, dynamic>? _dataMap;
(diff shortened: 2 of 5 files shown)


Afsal102:
cc: @Eimen2018

FilledStacks (changes requested):
Hey, 

This will make it so that the multi stream viewmodel can only be used with one type. This is not how it's used in production. We use this to get multiple different type of stream results that we want to rebuild on screen. Not just a single type. 

You could instead make the StreamData generic, that should help with your type safety per stream viewmodel you want.

Afsal102:
@FilledStacks oh, currently the key is string!, so even in production we still use one type!. am i missing something? 🤔 

> You could instead make the StreamData generic, that should help with your type safety per stream viewmodel you want

oh usually u will have only one type for Keys (currently its string(modified it to be user defined type)) at least thats what i use

>We use this to get multiple different type of stream results that we want to rebuild on screen. Not just a single type.

this MR doesnt modify anything related to the stream result it instead focuses on unique keys! that we pass in the stream  map
> that should help with your type safety per stream viewmodel you want

even after making this change u still have to do casting every where, once specific place would be onData callback

FilledStacks (changes requested):
My bad, I thought this was for the data type. I've never used a different type of key but that makes sense to have. Please increase the minor version number and set patch to 0 and then add the change into the changelog. 

Make double sure there's no breaking change.

Afsal102:
@FilledStacks haa got it, nothing breaks although i cant update the example haa it doesnt point to the local package path's

Afsal102:
@FilledStacks Bumped the package version!

FilledStacks:
Awesome. Please remove the +0 at the end of the changelog and the version number

Eimen2018:
This will be merged soon

Afsal102:
@Eimen2018 @FilledStacks thanks for making the change haa 🙇

FilledStacks:
This PR broke all existing stacked apps using multi stream viewmodel. I'm reverting until properly tested with backwards compatiblity.

Afsal102:
@FilledStacks ah my bad i think i should have made the commits concise that this would infact be a breaking change. 
A postmortem on why this will be a breaking change!
 - currently the keys are string, this MR will make them an object by default so all the overrides will show syntax errors, to fix this we just have to type the Multiple stream view model with <String>

As for the tests i think no additional tests are needed in my pov! As they wont provide value better than the existing ones( do let me know if u have any suggestions)

PR #594, opened 2022-05-02
Refactor/move the annotations from stacked package into a new package called stacked core

eb-so:
- [x]  Create a new package in the stacked repo called stacked_core
- [x]  Move all the annotation code from the stacked package to the stacked_core
- [x]  Replace all the stacked imports in the other packages with stack_core


packages/stacked_core/README.md
```diff
@@ -0,0 +1,39 @@
+<!--
```
FilledStacks: Please update to 


```md
# Stacked Core

This package contains all the code shared between our flutter and pure dart packages to avoid including flutter where it's not required.
```

FilledStacks (changes requested):
Only a single change required. Everything else should be good.

Can you please check the Example app still works using all the local package versions. 

I suspect that because the core annotations are not exported from the stacked package it'll break the example app and current apps using the annotations in production.

packages/stacked_core/README.md
eb-so: Done

FilledStacks (approved):
Awesome!

PR #595, opened 2022-05-02
Remove the export of Get package through stacked_services

eb-so:
- [x] Remove the export of Get package through stacked_services
- [x] Update the example to test if SnackPosition & SnackStyle still working


FilledStacks (approved):
🚀  Dope!

PR #599, opened 2022-05-04
Refactor/navigation service refactor

eb-so:
- Hide all the get package arguments in navigationservice
- Deprecate: transition, transitionClass, defaultTransition arguments
- Update the example app
- Replace deprecated documentaion


FilledStacks (approved):
Nice Work Ebrahim! This is awesome. 

Lets get this published!

PR #602, opened 2022-05-08
Cli Improvements

Eimen2018:
- Adds process service to handle command line commands
- Adds Colorized outputs 
- Fixes pubspec service not detecting the right file
- and a bunch of refactors and documentation

Preview 

https://user-images.githubusercontent.com/37356773/167316749-1787394a-a960-4cea-bd77-9b8e45bd544b.mov


FilledStacks (changes requested):
@Eimen2018 I like the log outputs. Something I noticed not is that the build output is included in our source control. We need to remove that. 

- Remove the build folder for the example project
- Remove any additional generated views like the testViews from the example when we commit. The playground should reflect the exact app that will be generated when stacked create app is run.

Eimen2018:
Update Preview -

https://user-images.githubusercontent.com/37356773/167444224-cdaa256c-776b-4af9-bda3-02b642c4fb3e.mp4

FilledStacks (changes requested):
This is great! The rest of the CLI updates can be done with the feature spec I shared. Rebase master into this branch and push again then i'll be ready for a merge.

FilledStacks (approved):

PR #605, opened 2022-05-12
Feature/add unit and integration tests to stacked_locator stacked_logger and refactor

eb-so:
- Remove ServicesConfig class and use the list of dependencies directly
- Revert FactoryParameter annotation to FactoryParam
- Setup locator should always have a future<void> return type
- Moved the logic from the stacked_locator_content to each dependency type `registerDependencies` function
- Add integration tests to locator_generator and logger_generator to check for compilation errors
- remove unused route samples
- Breaking change: replace stacked, stacked_annotations imports with stacked_core in locator generator
- Add the stacked_locator and stacked_logger to build.yaml local runner to run them from inside the package
- Remove flutter sdk dependency
- Add dependency_overrides for stacked_core
- Add meta package for visibleForTesting annotation
- Add Unit tests to all the possible scenarios of logger and locator generator content
-  add logger package to the dependency 
- replace kReleaseMode in logger_generator from flutter/foundation with `bool.fromEnvironment("dart.vm.product")` to get rid of flutter dependency


packages/stacked_generator/lib/src/generators/getit/dependency_config/factory_dependency.dart
```diff
@@ -0,0 +1,23 @@
...
+    Set<String>? environments,
+  }) : super(
+            import: import,
+            className: className,
+            abstractedImport: abstractedImport,
+            abstractedTypeClassName: abstractedTypeClassName,
+            environments: environments);
+
+  @override
+  String registerDependencies(String locatorName) {
```
FilledStacks, line 20: Nice

packages/stacked_generator/lib/src/generators/getit/dependency_config/factory_param_dependency.dart
```diff
@@ -0,0 +1,105 @@
...
+
+      constructorParamTypes.add('${paramConfig.type}');
+    });
+    if (factoryParamList.length == 1) {
+      constructorParamTypes.add('dynamic');
+    }
+
+    final constructerParams =
+        constructorParams.toString().replaceAll(RegExp('[{}]'), '');
+    final constructerParamtypes =
```
FilledStacks: This can be improved for what you need. I assume you only need to join the list with a `,` in between? If so you can use

```dart
final constructerParamtypes = constructorParamTypes.join(',')
```

packages/stacked_generator/lib/src/generators/getit/dependency_config/factory_param_dependency.dart
```diff
@@ -0,0 +1,105 @@
...
+        constructorParams.add('${paramConfig.name}:$getterName');
+      }
+
+      constructorParamTypes.add('${paramConfig.type}');
+    });
+    if (factoryParamList.length == 1) {
+      constructorParamTypes.add('dynamic');
+    }
+
+    final constructerParams =
```
FilledStacks: The same join functionality can be applied here.

packages/stacked_generator/lib/src/generators/getit/dependency_config/dependency_config.dart
```diff
@@ -1,46 +1,29 @@
...
     required this.import,
     required this.className,
     this.abstractedImport,
+    this.environments,
     this.abstractedTypeClassName,
-    this.presolveFunction,
-    this.resolveFunction,
   });
 
+  String registerDependencies(String locatorName);
```
FilledStacks, line 25: This override abstraction is awesome! Great choice

packages/stacked_generator/lib/src/generators/getit/stacked_locator_content_generator.dart
```diff
@@ -1,35 +1,30 @@
...
     newLine();
     writeLine('final $locatorName = StackedLocator.instance;');
     newLine();
 
-    final hasPresolve = services
-        .any((service) => service.type == DependencyType.PresolvedSingleton);
-
     writeLine(
-        '${hasPresolve ? 'Future' : 'void'} $locatorSetupName ({String? environment , EnvironmentFilter? environmentFilter}) ${hasPresolve ? 'async' : ''} {');
+        'Future<void> $locatorSetupName ({String? environment , EnvironmentFilter? environmentFilter}) async {');
```
FilledStacks, line 27: If we have no presolve then this doesn't need to be a Future. Which is why I had the original check there. 

If it's always a Future that means the main function will always have to be turned into a Future?

packages/stacked_generator/test/getit/unit_test/factory_test.dart
```diff
@@ -0,0 +1,39 @@
...
+      expect(
+          () => factoryLocator<DumpService>(), throwsA(isA<AssertionError>()));
+    });
+    test('When register factory, Should returnsNormally', () async {
+      setupFactoryLocator();
+
+      expect(() => factoryLocator<DumpService>(), returnsNormally);
+    });
+    test(
+        'When register factory with type, Should call it by the type not the implementation',
```
FilledStacks, line 23: Nice set of tests!

FilledStacks (commented):
Awesome PR! All the tests are passing. Some failed at first until I ran a dart pub upgrade. 

But now it's all passing. I'm very happy with this PR. It can be merged (after changes) and the same can be applied to other generators where required.

packages/stacked_generator/lib/src/generators/getit/stacked_locator_content_generator.dart
eb-so: I think it's more consistent, and there is no harm of doing it

packages/stacked_generator/lib/src/generators/getit/dependency_config/factory_param_dependency.dart
eb-so: I didn't write this code I just moved it to another file, I will clean it up 👍

packages/stacked_generator/lib/src/generators/getit/stacked_locator_content_generator.dart
FilledStacks: @eb-so we'll have to update the stacked readme to direct users to always update the main from `void` to `Future<void>` and call `await setupLocator()` instead of only `setupLocator`

FilledStacks (approved):
Happy with this PR! Well done, much needed for the project

packages/stacked_generator/lib/src/generators/getit/stacked_locator_content_generator.dart
eb-so: I'll make a ticket for that

FilledStacks:
@eb-so let me know what the hold up is of merging this in.

eb-so:
@FilledStacks Nothing, You can merge it 💯

PR #607, opened 2022-05-13
feat: Flutter v3 compatibility - Remove operand of null-aware operation

ghost:
Hello @FilledStacks, 
I fixed all errors when launching the project after migrations to Flutter v3.
Let me know if this helps and how to improve if needed. 

Have a nice day!

_Gerfaut_


ghost:
Maybe this should not be fixed as is, but with support for 3.0 and earlier version, using the _ambiguate solution provided by the Flutter team: https://docs.flutter.dev/development/tools/sdk/release-notes/release-notes-3.0.0#your-code.
What do you think?

packages/stacked_themes/lib/src/theme_manager.dart
```diff
@@ -50,7 +50,7 @@ class ThemeManager {
   /// Returns the current selected theme mode
   ThemeMode? get selectedThemeMode => _selectedThemeMode;
 
-  ThemeModel get initalTheme => _initialTheme!;
```
pratikbaid3, line 53: Any specific reason for this change. It just looks like a change in the variable name. For projects using linting, this will cause issues since the spelling is wrong.

FilledStacks:
@Gerfaut thanks! I'm checking out the fix and your code now.

FilledStacks:
@Gerfaut I like the ambiguate solution. Please use that one and add the link to those docs in a comment. This way we'll know to remove it once everything is moved over to Flutter 3.

ghost:
Hey @FilledStacks,
Thanks for feedback! I just pushed the changes and I hope it fit your expectations 🙏 
Let me know if it isn't the case! 
Have a nice day 🌞 

_Gerfaut_

FilledStacks:
It's gonna be published today

mugikhan:
@FilledStacks This warning still appears in version 0.3.8+1 of the stacked_themes package. Is there anyway you can update that package?

PR #608, opened 2022-05-13
fix: :bug: texteditingcontroller losing state

Eimen2018:
- Fixes TextEditingController Losing State on hot-reload
- Adds Some Unit Test to make sure it generates the right output
- Adds FormUtilService

### DIFF
--- a/packages/stacked/example/pubspec.yaml
+++ b/packages/stacked/example/pubspec.yaml
@@ -49,7 +49,7 @@ dev_dependencies:
   # dependency injection
   build_runner:
   stacked_generator:
-    path: ../../stacked_generator
+    # path: ../../stacked_generator
 
 # For information on the generic Dart part of this file, see the
 # following page: https://dart.dev/tools/pub/pubspec
--- a/packages/stacked/example/pubspec.yaml
+++ b/packages/stacked/example/pubspec.yaml
@@ -49,7 +49,7 @@ dev_dependencies:
   # dependency injection
   build_runner:
   stacked_generator:
-    # path: ../../stacked_generator
+    path: ../../stacked_generator
 
 # For information on the generic Dart part of this file, see the
 # following page: https://dart.dev/tools/pub/pubspec
--- a/packages/stacked_generator/lib/src/generators/forms/stacked_form_content_generator.dart
+++ b/packages/stacked_generator/lib/src/generators/forms/stacked_form_content_generator.dart
@@ -1,41 +1,41 @@
-import 'package:stacked_generator/src/generators/forms/form_util_service.dart';
+import 'package:stacked_generator/src/generators/forms/form_generator_util.dart';
 import 'package:stacked_generator/src/generators/forms/form_view_config.dart';
 
 /// A generator that creates the form code based on the configs passed in
-class StackedFormContentGenerator extends FormUtilService {
-  final FormViewConfig formViewConfig;
-  StackedFormContentGenerator(this.formViewConfig)
-      : super(formViewConfig: formViewConfig);
+class StackedFormContentGenerator {
+  final FormViewConfig _formViewConfig;
+  StackedFormContentGenerator(this._formViewConfig) : super();
+
+  late FormGeneratorUtil _util;
 
   String generate() {
-    writeLine("// ignore_for_file: public_member_api_docs");
+    _util = FormGeneratorUtil(formViewConfig: _formViewConfig);
+    _util.writeLine("// ignore_for_file: public_member_api_docs");
 
-    // TODO: All of these functions can be moved into a separate util class or service
-    // and can then be unit tested to avoid simple mistakes. BUT, that's for later.
-    generateImports();
-    generateValueMapKeys();
-    generateDropdownItemsMap();
-    generateTextEditingControllerItemsMap();
+    _util.generateImports();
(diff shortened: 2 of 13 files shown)


eb-so (commented):
ignore this comment

packages/stacked/example/pubspec.yaml
```diff
@@ -49,7 +49,7 @@ dev_dependencies:
   # dependency injection
   build_runner:
   stacked_generator:
-    # path: ../../stacked_generator
+    path: ../../stacked_generator
```
eb-so: Revert this back or use dependency_override to use path in development and the actual package in production

packages/stacked_generator/lib/src/generators/forms/stacked_form_content_generator.dart
```diff
@@ -1,262 +1,41 @@
...
-import 'package:stacked_generator/src/generators/forms/field_config.dart';
+import 'package:stacked_generator/src/generators/forms/form_util_service.dart';
 import 'package:stacked_generator/src/generators/forms/form_view_config.dart';
 
 /// A generator that creates the form code based on the configs passed in
-class StackedFormContentGenerator extends BaseGenerator {
-  final FormViewConfig _formViewConfig;
-
-  StackedFormContentGenerator(this._formViewConfig);
+class StackedFormContentGenerator extends FormUtilService {
```
eb-so: I don't like this cause it will expose all the methods of FormUtilService while we want StackedFormContentGenerator to have only one public method which is `generate`

packages/stacked_generator/lib/src/generators/forms/stacked_form_content_generator.dart
```diff
@@ -1,262 +1,41 @@
...
-    _generateImports();
-    _generateValueMapKeys(fields);
-    _generateDropdownItemsMap(fields.onlyDropdownFieldConfigs);
+    generateImports();
+    generateValueMapKeys();
+    generateDropdownItemsMap();
+    generateTextEditingControllerItemsMap();
     _generateFormMixin();
-    _generateFormViewModelExtensions(fields);
+    _generateFormViewModelExtensions();
```
eb-so: Mixing the private and public method is a bad practice 
I prefere if you revert to extend the `BaseGenerator` and use the `FormUtilService` as a helper class

packages/stacked_generator/test/form_generator/form_generator_test.dart
```diff
@@ -0,0 +1,94 @@
...
+import 'package:stacked_generator/src/generators/forms/form_util_service.dart';
+import 'package:stacked_generator/src/generators/forms/form_view_config.dart';
+import 'package:test/test.dart';
+
+import 'constant_test_helper.dart';
+
+void main() {
+  group('FormGeneratorTest -', () {
+    group('generateImport -', () {
+      test('when called should generate imports', () {
```
eb-so: the tests naming is very good and consice, I like it

packages/stacked_generator/test/form_generator/form_generator_test.dart
```diff
@@ -0,0 +1,94 @@
...
+        FormUtilService utilService = FormUtilService(
+          formViewConfig: FormViewConfig(
+            viewName: 'Test',
+            fields: [],
+          ),
+        );
+        utilService.generateImports();
```
eb-so: the class name is FormGeneratorTest but you're acctually testing FormUtilService 
the fix I talked about in the previous comments will fix your tests too

eb-so (changes requested):

packages/stacked_generator/lib/src/generators/forms/stacked_form_content_generator.dart
```diff
@@ -1,262 +1,41 @@
...
-  void _generateTextEdittingControllersForTextFields(
-      List<TextFieldConfig> fields) {
-    for (final field in fields) {
-      writeLine(
-          'final TextEditingController ${_getControllerName(field)} = TextEditingController(${_getControllerInitialValue(field)});');
-    }
+  void _generateFormViewModelExtensions() {
+    _util.generateFormViewModelExtensionForGetters();
+    _util.generateFormViewModelExtensionForMethods();
   }
```
eb-so: Why don't you move this to the util class too

packages/stacked_generator/lib/src/generators/forms/stacked_form_content_generator.dart
```diff
@@ -1,262 +1,41 @@
...
-import 'package:stacked_generator/src/generators/base_generator.dart';
-import 'package:stacked_generator/src/generators/forms/field_config.dart';
+import 'package:stacked_generator/src/generators/forms/form_generator_util.dart';
 import 'package:stacked_generator/src/generators/forms/form_view_config.dart';
 
 /// A generator that creates the form code based on the configs passed in
-class StackedFormContentGenerator extends BaseGenerator {
+class StackedFormContentGenerator {
   final FormViewConfig _formViewConfig;
+  StackedFormContentGenerator(this._formViewConfig) : super();
```
eb-so: remove unnessary super call

eb-so:
Nice work @Eimen2018 , LGTM ✅

PR #494, opened 2021-08-16
add move in left transition

saileshbro:

PR #609, opened 2022-05-16
feature/add move in left route transtion

eb-so:
Replaces #494


Eimen2018:
LGTM 🔥

FilledStacks (changes requested):
Change the version number to 2.3.7 instead.

For versioning a good guide is.
- When we collect a number of minor improvements and features we bump the major version
- When we implement new functionality we increase the minor version (second number)
- If we fix something or add to an existing implementation .i.e. add a new transition we increase the patch number (last digit)
- If no code changes or we only update configuration and the functionality of everything is still the same we add a +version

eb-so:
I'll save this note for future releases 👍

PR #611, opened 2022-05-17
Improve Console Writes

Eimen2018:
- Adds warn, error, success, info colored logs
- Adds modified name for verbose message on file modification
- Adds process service unit test
- Adds Additional warning logs on file or file identifier missing
- some refactors


eb-so:
LGTM

PR #612, opened 2022-05-18
Use dependency overrides instead of comment out/in the local  dependency

Eimen2018:
- Adds all local package to dependency overides to use path on dev

- Fixes flutter V3 introduced bugs

### DIFF
--- a/packages/stacked/example/pubspec.yaml
+++ b/packages/stacked/example/pubspec.yaml
@@ -32,12 +32,12 @@ dependencies:
   stacked_themes:
   stacked_crashlytics:
   
-  flutter_hooks:
+  flutter_hooks: ^0.18.4
   # navigation
-  get:
+  get: ^4.6.3
 
   # logging
-  logger:
+  logger: ^1.1.0
 
 dependency_overrides: 
   stacked:
--- a/packages/stacked/example/pubspec.yaml
+++ b/packages/stacked/example/pubspec.yaml
@@ -36,9 +36,6 @@ dependencies:
   # navigation
   get: ^4.6.3
 
-  # logging
-  logger: ^1.1.0
-
 dependency_overrides: 
   stacked:
       path: ../
--- a/packages/stacked/example/pubspec.yaml
+++ b/packages/stacked/example/pubspec.yaml
@@ -27,20 +27,32 @@ dependencies:
 
   # state management
   stacked:
-    path: ../
   stacked_services:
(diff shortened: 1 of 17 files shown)


packages/stacked_localisation/stacked_localisation_generator/pubspec.yaml
```diff
@@ -9,7 +9,7 @@ environment:
 dependencies:
   build: ^1.3.0
   source_gen: ^0.9.6
-  analyzer: ^0.39.17
+  analyzer:
```
eb-so: I think it's a bad practice to depend on a package without a version, If you tried to publish the package with `flutter pub publish --dry-run` it will show a warning message about this
https://medium.com/cashify-engineering/flutter-dependencies-resolution-2faecdbfce03

packages/stacked_themes/example/pubspec.yaml
```diff
@@ -29,9 +29,14 @@ dependencies:
   cupertino_icons: ^0.1.3
   stacked:
   stacked_themes:
-    path: ../
   get_it:
```
eb-so: Add version here too

packages/stacked/example/pubspec.yaml
```diff
@@ -27,21 +27,33 @@ dependencies:
...
   stacked_hooks:
-  flutter_hooks:
   stacked_themes:
-    path: ../../stacked_themes/
   stacked_crashlytics:
-    path: ../../stacked_crashlytics/
+  
+  flutter_hooks:
   # navigation
   get:
```
eb-so: same

packages/stacked/example/pubspec.yaml
```diff
@@ -27,21 +27,33 @@ dependencies:
...
 
   # logging
   logger:
```
eb-so: We don't need the `logger` any more cause I added it to the stacked_generator dependencies

eb-so (commented):
Some minor changes

eb-so:
LGTM ✅

ferrarafer (approved):
LGTM

FilledStacks (approved):
LGTM. 

This will have to be commented out before release

eb-so:
@FilledStacks from the [docs](https://dart.dev/tools/pub/dependencies#dependency-overrides)
Screen Shot 2022-05-19 at 4 11 05 PM

PR #614, opened 2022-05-18
Refactor/add unit tests and refactor route generator

eb-so:


packages/stacked_generator/lib/src/generators/exceptions/invalid_generator_input_exception.dart
```diff
@@ -10,3 +10,10 @@ class InvalidGeneratorInputException implements Exception {
   String toString() =>
       'InvalidGeneratorInputException: message: $message\n todo: $todo\n element: $element';
 }
+
+class ExceptionMessages {
+  static const RouteClassNameShouldnotBeNull = 'className should not be null';
```
ferrarafer, line 15: Typo
RouteClassNameShould**n**otBeNull -> RouteClassNameShould**N**otBeNull

packages/stacked_generator/lib/src/generators/router/router_class_generator.dart
```diff
@@ -129,6 +131,8 @@ class RouterClassGenerator extends BaseGenerator {
 
     var routesMap = <String, RouteConfig>{};
     routerConfig.routes?.forEach((route) {
+      throwIf(route.className == null,
+          ExceptionMessages.RouteClassNameShouldnotBeNull);
```
ferrarafer: Typo
RouteClassNameShould**n**otBeNull -> RouteClassNameShould**N**otBeNull

packages/stacked_generator/lib/src/generators/router/router_class_generator.dart
```diff
@@ -145,10 +149,11 @@ class RouterClassGenerator extends BaseGenerator {
 
     if (r.parameters?.isNotEmpty == true) {
```
ferrarafer: I never saw this shortcut, how does it work? I assume is the same as doing `r.parameters?.isNotEmpty ?? true`,  right?

ferrarafer (commented):
LGTM, just typo fixes

packages/stacked_generator/lib/src/generators/router/router_class_generator.dart
eb-so: Me neither, this is 15 months old code, but I guess it should read like this 
1- if isNotEmpty is null then `null==true` which is false
2- if isNotEmpty is not null then it should execute normally

equals to
`r.parameters?.isNotEmpty ?? false`

FilledStacks:
@eb-so I'm merging this in!

eb-so:
I didn't finish up the work but that's ok I can create a new PR for the rest

PR #623, opened 2022-05-27
Fix/stacked generator post-refactor

eb-so:
- Stacked_generator:replace the RouterConfig field with routeNamePrefex in RouteConfigResolver
- Fix nested route resolve in RouterConfigResolver
- Fix some imports


ferrarafer (approved):
LGTM

PR #626, opened 2022-05-29
Chore/add integration test to stacked example app

isaacadariku:
This PR adds Integration test to stacked example app

- [x] Update existing example app to latest stacked setup
- [x] Fix all flutter analyzer issue
- [x] Connect views to each other for end to end test
- [x] Adds integration test
- [x] Create integration test workflow to run on Github Action using Chrome Driver

<a href="https://www.loom.com/share/86aa1673c8004b93b7c5efb73d41ed86">
    <p>Stacked app end to end testing Emulator demo - Watch Video</p>
    
  </a>

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -20,7 +20,6 @@ jobs:
 
       - name: Setup chrome browser
         uses: browser-actions/setup-chrome@latest
-        run: chrome --102.0.5005.61
 
       - name: Setup chrome driver
         uses: nanasess/setup-chromedriver@v1
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -4,25 +4,28 @@ on: [push, pull_request]
 
 jobs:
   integration_tests:
+    name: Run stacked app integration tests
     runs-on: ubuntu-latest
-
-
-env:
-  flutter_channel: "stable"
-
+    if: github.event.pull_request.draft == false
     steps:
+      - name: Setup Flutter
       - uses: actions/checkout@v2
       - uses: subosito/flutter-action@v1
         with:
-          channel: ${{ env.flutter_channel }}
+          channel: flutter_channel: "stable"
 
+      - name: Setup chrome driver
       - uses: actions/checkout@v2
       - uses: nanasess/setup-chromedriver@v1
         with:
           chromedriver_version: "102.0.5005.61"
+
+      - name: Install dependencies
       - run: flutter pub get
-      - run: chromedriver --port=4444 &
+
+      - name: Run integration tests
       - run: cd packages/stacked/example
+      - run: chromedriver --port=4444 &
       - run: |
           flutter drive \
             --driver=test_driver/integration_test.dart \
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -8,14 +8,14 @@ jobs:
     runs-on: ubuntu-latest
     if: github.event.pull_request.draft == false
     steps:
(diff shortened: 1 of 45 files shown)


.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,35 @@
...
+        with:
+          chromedriver-version: "101.0.4951.41"
+
+      - name: Install dependencies
+        run: flutter pub get
+
+      - name: Launch web driver & run tests
+        run: |
+          chromedriver --port=4444 & 
+          flutter drive \
```
FilledStacks, line 32: I'm curios to know why we're using flutter driver here if the integration_tests package is used to run the tests locally?

packages/stacked/example/integration_test/app_test.dart
```diff
@@ -0,0 +1,295 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:integration_test/integration_test.dart';
+import 'package:new_architecture/main.dart' as app;
+
+void main() {
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
+  group('end-to-end test', () {
```
FilledStacks: I think these names can be more descriptive. I also don't like the fact that it's one large test, BUT, it does get the job done that we want for now so I'm happy with that. Only changes that I'd want is:

Change group name to "Integration Test"
Change test name to "Ensure all stacked integrations still work as expected"

FilledStacks (changes requested):
Nice PR Isaac. Very happy with it. This will be a major time saver for anyone that's making changes to Stacked. 

There's only some name changes requested. Could you please also add a new file CONTRIBUTION.md in the root of the stacked package and in there add:

## Running Tests

To ensure that everything still works run the integration tests in the example folder. Use the command

```
[Add the command to run the tests from the root package here]
```

By the time you see this, if it's 30 May then the contribution file will already exist because I am making it today after getting all the docs ready.

packages/stacked/example/integration_test/app_test.dart
isaacadariku: 😶‍🌫️ Yeah. thanks! I will work on regrouping it to it's specific test. I was just ensuring end to end works for now 😄

.github/workflows/integration-tests.yml
isaacadariku: According to the doc the `integration_tests` package  is compatible with the `flutter drive` command and the command above help us run the test on the web with the help of the chrome driver. My aim was to setup a CI that runs on the web and don't want us to depend on the contributor local test alone.

Screenshot 2022-05-29 at 11 30 34

This is also the CI on my fork branch
image

For the new commits, incase a contributor just want to use their terminal for testing, it uses the `flutter driver` command.
Here is demo for it 
<a href="https://www.loom.com/share/94e4e5e49dfa48389188502d0f63dfcc">
    <p>Running with chrome driver on CLI - Watch Video</p>
    
  </a>

CONTRIBUTION.MD
```diff
@@ -0,0 +1,29 @@
...
+
+Use the command:
+
+```
+flutter test integration_test/app_test.dart
+```
+
+If multiple devices was found, select the device you want to run the tests on.
+
+## **Running using command line**
```
ferrarafer: What do you mean with this? The action above is also using command line, maybe we need to change this to something more descriptive.

isaacadariku: Thanks for pointing it out. I have made the changes.

eb-so:
LGTM 🎉 🎉

PR #628, opened 2022-05-30
Refactor/update stacked generator to flutter 3.0

eb-so:
- extract the RouteConfigFactory to a seperate class
- convert all RouteConfig supclasses to use super paremeters
- refactor route config resolver


FilledStacks (approved):
LGTM

PR #629, opened 2022-05-30
add build github action

eb-so:
- Add stacked, stacked_generator unit tests to run on every push, PR, or manually
- Change stacked example app name back to 'example' from new_architecture
- Fix analyzer issues in the router annotations


.github/workflows/build.yaml
```diff
@@ -0,0 +1,90 @@
+name: Build
+
+on:
+  push:
```
FilledStacks, line 4: Can we make this push on master only please

FilledStacks (changes requested):
Good job on that rename catch. 

One change requested for the CI work.

FilledStacks (approved):
LGTM

PR #632, opened 2022-05-30
Fixes Disposed TextEditingController when navigating away

### LINKED ISSUE #627: TextEdittingController can no longer be used
After upgrading when I use the Navigation service to navigate to route A, and then I use `.back()` to return to the home page, I get an exception when navigating to route A again.

```
A TextEditingController was used after being disposed. Once you have called dispose() on a TextEditingController, it can no longer be used.
```

This was introduced with #608

Eimen2018:
- fixes #627


ferrarafer (commented):
LGTM

eb-so:
LGTM

FilledStacks (approved):
I'm Happy with the update. This will ensure when a controller is requested one will be created if it doesn't exist and if someone makes use of the dispose they will be disposed properly and then cleared from the map. 

Much better.

PR #634, opened 2022-05-30
Remove route guard from router annotation

### LINKED ISSUE #449: RouteGuards are never reached
Hi!

First of all, awesome package! But I am having issues implementing RouteGuards.

Following the documentation example, here is my main.dart:

```dart
class MyApp extends StatelessWidget {
  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      title: 'Disco',
      navigatorKey: StackedService.navigatorKey,
      onGenerateRoute: StackedRouter().onGenerateRoute,
      theme: ThemeData(
        appBarTheme: AppBarTheme(color: Colors.white),
        fontFamily: 'Rubik',
        primarySwatch: MaterialColor(0xFF3E9FC9, primaryColor),
        textTheme: Theme.of(context).textTheme.apply(
              fontFamily: 'Rubik',
              bodyColor: Color.fromRGBO(23, 40, 56, 1),
              displayColor: Color.fromRGBO(23, 40, 56, 1),
            ),
        visualDensity: VisualDensity.adaptivePlatformDensity,
      ),
    );
  }
}
```

Here is my route with its guard:

```dart
AdaptiveRoute(
  page: AuthenticationView,
  name: 'Authentication',
  guards: [SignedOutOnlyGuard],
)
```

Here is the generated RouteDef inside app.router.dart:

```dart
RouteDef(Routes.Authentication,
        page: AuthenticationView, guards: [SignedOutOnlyGuard])
```

And finally, here is my RouteGuard

```dart
class SignedOutOnlyGuard extends RouteGuard {
  final AuthUseCase _auth = GetIt.instance<AuthUseCase>();

  @override
  Future<bool> canNavigate(
    ExtendedNavigatorState<RouterBase?> navigator,
    String routeName,
    Object? arguments,
  ) async {
    switch (_auth.state) {
      case AuthState.LOADING:
        navigator.replace(Routes.SplashScreen);
        return false;
      case AuthState.ERROR:
        await _auth.signOut();
        return true;
      case AuthState.AUTHENTICATED:
        navigator.replace(Routes.Feed);
        return false;
      default:
        return false;
    }
  }
}
```

The problem is that this canNavigate method is never called and the user can navigate to it regardless of the guard's logic. Are there any other steps that I am missing? The documentation doesn't mentioned guards

I am using stacked: 2.2.2

eb-so:
- Remove `guard` field which is only required when working with flutter web(Not a high priority target for stacked for now)
- Closes #449


ferrarafer (commented):
LGTM

FilledStacks (approved):
LGTM. We should always try and remove half baked features.

PR #635, opened 2022-05-31
Feature/stongly-typed-navigation

Eimen2018:
- Adds strongly typed navigation 
- Adds custom transition option for custom routes


packages/stacked/example/lib/app/app.dart
```diff
@@ -19,7 +19,7 @@ import 'package:stacked_themes/stacked_themes.dart';
     MaterialRoute(page: HomeView, initial: true),
     MaterialRoute(page: BottomNavExample),
     MaterialRoute(page: StreamCounterView),
-    CupertinoRoute<Map<String, List<String>>>(page: DetailsView),
+    CupertinoRoute<Map<String, List<String>>?>(page: DetailsView),
```
eb-so, line 22: To activate the type inference we have to add it here right?

If the above is right what happen when we change one of the class parameters without adjusting the type here?

packages/stacked_generator/lib/src/generators/getit/stacked_locator_content_generator.dart
```diff
@@ -15,7 +15,7 @@ class StackedLocatorContentGenerator extends BaseGenerator {
   });
   @override
   String generate() {
-    writeLine("// ignore_for_file: public_member_api_docs");
+    writeLine("// ignore_for_file: public_member_api_docs, unused_import");
```
eb-so: Why did you add ", unused_import"

packages/stacked_services/lib/src/navigation/navigation_service.dart
```diff
@@ -187,10 +187,11 @@ class NavigationService {
     int? id,
     bool preventDuplicates = true,
     Map<String, String>? parameters,
+    RouteTransitionsBuilder? transition,
```
eb-so, line 190: I think this is a fix for adding transtion to navigation with named route right?

packages/stacked/example/lib/app/app.dart
Eimen2018: This is the expected return type when navigating back. if the user presses the back button it would be null causing a null issue. that's why i added the question mark... to indicate it might be null

packages/stacked_generator/lib/src/generators/getit/stacked_locator_content_generator.dart
Eimen2018: I added it before the imports was fixed I will remove it

packages/stacked_services/lib/src/navigation/navigation_service.dart
Eimen2018: Yup! 👍

eb-so:
@Eimen2018 Awesome work Ayman you completed this feature in a short period of time and also solved another issue"Custom transitions when using named routes." which is Great 🎉 
My only request is to create a new PR for the "Custom transitions when using named routes." feature and merge it first then you can complete your work on the strong type navigation

FilledStacks (approved):
The implementation looks good! Get those Unit tests fixed up and this will be ready to push out.

FilledStacks:
I'm approving this and merging it int. There's been no objections.

PR #636, opened 2022-06-01
Added Delete Service Command

BlueDev5:
I've added the delete service command, and to use it we can run
`stacked delete service name_of_the_service`
and to achieve this, I added a function called revertTemplate() to the template service, which takes all the same args as renderTemplate() from the template service. Instead of rendering the template, it goes ahead and deletes all the template files which are created, and reverts all the modifications made to existing files, while rendering that template. So now, when deleting a view or service, or theoretically any template, you can use the renderTemplate() to create the template and revertTemplate() to revert all those changes made during renderTemplate().


FilledStacks:
Awesome. I will review this when I'm back at my desk. Thanks for the contribution

packages/stacked_tools/lib/src/commands/delete/delete_view_command.dart
```diff
@@ -36,8 +33,7 @@ class DeleteViewCommand extends Command with ProjectStructureValidator {
     final outputPath = argResults!.rest.length > 1 ? argResults!.rest[1] : null;
```
FilledStacks: All of this code here is repeated. We always want to aim for DRY code (Don't Repeat Yourself). This can be extracted into an abstract class that has a run implementation and calls into a required implemeting function called purge that takes in the output path.

packages/stacked_tools/lib/src/services/file_service.dart
```diff
@@ -94,7 +94,7 @@ class FileService {
     await writeFile(
       file: File(filePath),
       fileContent: fileLines.join('\n'),
-      type: FileModificationType.Modify,
+      type: FileModificationType.ApplyModification,
```
FilledStacks, line 97: Nice naming

packages/stacked_tools/lib/src/services/process_service.dart
```diff
@@ -14,7 +14,7 @@ class ProcessService {
   /// Args:
   ///   appName (String): The name of the app that's going to be create.
   Future<void> runCreateApp({required String appName}) async {
-    await _runProcessAndLogOutput(
```
FilledStacks: Please change this back to a private function. None of the code should be running process commands directly. Put it behing a function similar to createApp so that we can manage all that code in one place.

packages/stacked_tools/lib/src/services/template_service.dart
```diff
@@ -339,4 +467,51 @@ class TemplateService {
       '$renderedTemplate\n$modificationIdentifier',
     );
   }
+
+  /// It reverts the modification applied to a file by [modificationTemplate]
+  Future<String> templateWithoutModifiedFileContent({
```
FilledStacks: No unit tests has been added for this functionality. Please add tests to confirm that all this works objectively.

FilledStacks (changes requested):
I really appreicate the PR for this functionality. There are 2 things that are missing:
- Unit tests (objectively prove that your tests do what you think it does)
- Actual testing. There's a bug in the code. 

## Steps
1. Create a service called serviceOne
2. Create a service called serviceTwo
3. Run `stacked delete service serviceOne` command 

**Expected:** ServiceOnce class is removed and all the code in the test_helpers is reverted for that service only

**Actual:** The the first service entries in the test_helpers is still there and breaks compilation because only the files were removed. 

Apply some of the refactors and then focus some time on the tests to make sure we can prove your code is working as expected.

BlueDev5:
The bug is now fixed, now I'll start with resolving all the changes you've requested

packages/stacked_tools/lib/src/commands/delete/delete_view_command.dart
BlueDev5: I've extracted that function as purgeTemplate, instead of what you told me because I realized the only difference in both commands was just the template name, everything else was just a copy. and it seemed to best suit the template service, but now the template service file is nearly 550 lines long. So I think that I should separate the file into 3 parts, namely, renderTemplateService, revertTemplateService, and templateServiceUtils.
After this the create commands will depend on renderTemplateService, delete commands will depend on revertTemplateService and both of these new services will depend on templateServiceUtils.

BlueDev5:
Just wanted to let you know, the currently written tests for template Service seem to be broken.
all of them result in the same error:
```
'package:get_it/get_it_impl.dart': Failed assertion: line 372 pos 7: 'instanceFactory != null': Object/factory with  type ProcessService is not registered inside GetIt. 
(Did you accidentally do GetIt sl=GetIt.instance(); instead of GetIt sl=GetIt.instance;
Did you forget to register it?)
dart:core                                                       _AssertionError._throwNew
package:get_it/get_it_impl.dart 372:7                           _GetItImplementation._findFactoryByNameAndType
package:get_it/get_it_impl.dart 393:29                          _GetItImplementation.get
package:get_it/get_it_impl.dart 430:12                          _GetItImplementation.call
package:stacked_tools/src/services/template_service.dart 30:34  new TemplateService
test/services_tests/template_service_test.dart 13:34            _getService
test/services_tests/template_service_test.dart 35:33            main.<fn>.<fn>.<fn>
```
So until they don't get fixed (by me or anyone else, I think it's not possible for me to write those tests)

packages/stacked_tools/lib/src/services/file_service.dart
BlueDev5: Thanks! I did that cause it wasn't easy to determine when the logs should say reverted and when they should say applied modification with that old enum

packages/stacked_tools/lib/src/services/process_service.dart
BlueDev5: Extracted a new function "formatFile" and made this function private

packages/stacked_tools/lib/src/services/template_service.dart
BlueDev5: I've added 24 tests, they are organised as followed:

4 tests for single line templates
4 tests for multi-line templates, which also change on running dart or flutter format on them, for ex:
   ```
int getSomeInt(){
return 1;
}
```
This on getting formatted becomes:
```
int getSomeInt () {
  return 1;
}
```
and finally 4 for multi-line templates that do not get changed even if you run dart or flutter format on them.

12 tests where the template doesn't use any variable like {{serviceName}}.
12 tests where the template does use variables like {{serviceName}}.

Both types have the same distribution as I mentioned above, i.e. 4 single line, 4 multi-line (change on getting formatted), and 4 multi-line (don't change on getting formatted)

File: packages/stacked_tools/lib/src/commands/create/create_app_command.dart [hunk 5]
```diff
@@ -4,11 +4,11 @@ import 'package:args/command_runner.dart';
 import 'package:stacked_tools/src/locator.dart';
 import 'package:stacked_tools/src/services/colorized_log_service.dart';
 import 'package:stacked_tools/src/services/process_service.dart';
-import 'package:stacked_tools/src/services/template_service.dart';
+import 'package:stacked_tools/src/services/render_template_service.dart';
 import 'package:stacked_tools/src/templates/template_constants.dart';
 
 class CreateAppCommand extends Command {
-  final _templateService = locator<TemplateService>();
```
FilledStacks, line 11: service variable name should match the service class name.

packages/stacked_tools/lib/src/commands/create/create_service_command.dart
```diff
@@ -4,13 +4,13 @@ import 'package:stacked_tools/src/locator.dart';
...
 import 'package:stacked_tools/src/services/process_service.dart';
-import 'package:stacked_tools/src/services/template_service.dart';
+import 'package:stacked_tools/src/services/render_template_service.dart';
 import 'package:stacked_tools/src/templates/template_constants.dart';
 
 import '../../services/pubspec_service.dart';
 
 class CreateServiceCommand extends Command with ProjectStructureValidator {
-  final _templateService = locator<TemplateService>();
+  final _templateService = locator<RenderTemplateService>();
```
FilledStacks: service variable name should match the service class name.

packages/stacked_tools/lib/src/commands/create/create_view_command.dart
```diff
@@ -5,11 +5,11 @@ import 'package:stacked_tools/src/constants/message_constants.dart';
 import 'package:stacked_tools/src/mixins/project_structure_validator_mixin.dart';
 import 'package:stacked_tools/src/services/process_service.dart';
 import 'package:stacked_tools/src/services/pubspec_service.dart';
-import 'package:stacked_tools/src/services/template_service.dart';
+import 'package:stacked_tools/src/services/render_template_service.dart';
 import 'package:stacked_tools/src/templates/template_constants.dart';
 
 class CreateViewCommand extends Command with ProjectStructureValidator {
-  final _templateService = locator<TemplateService>();
+  final _templateService = locator<RenderTemplateService>();
```
FilledStacks: service variable name should match the service class name.

packages/stacked_tools/lib/src/commands/delete/delete_service_command.dart
```diff
@@ -0,0 +1,32 @@
+import 'package:args/command_runner.dart';
+import 'package:stacked_tools/src/constants/command_constants.dart';
+import 'package:stacked_tools/src/constants/message_constants.dart';
+import 'package:stacked_tools/src/locator.dart';
+import 'package:stacked_tools/src/mixins/project_structure_validator_mixin.dart';
+import 'package:stacked_tools/src/services/revert_template_service.dart';
+import 'package:stacked_tools/src/templates/template_constants.dart';
+
+class DeleteServiceCommand extends Command with ProjectStructureValidator {
+  final _templateService = locator<RevertTemplateService>();
```
FilledStacks: service variable name should match the service class name.

packages/stacked_tools/lib/src/commands/delete/delete_view_command.dart
```diff
@@ -0,0 +1,31 @@
+import 'package:args/command_runner.dart';
+import 'package:stacked_tools/src/constants/command_constants.dart';
+import 'package:stacked_tools/src/constants/message_constants.dart';
+import 'package:stacked_tools/src/locator.dart';
+import 'package:stacked_tools/src/mixins/project_structure_validator_mixin.dart';
+import 'package:stacked_tools/src/services/revert_template_service.dart';
+import 'package:stacked_tools/src/templates/template_constants.dart';
+
+class DeleteViewCommand extends Command with ProjectStructureValidator {
+  final _templateService = locator<RevertTemplateService>();
```
FilledStacks: service variable name should match the service class name.

packages/stacked_tools/lib/src/services/revert_template_service.dart
```diff
@@ -0,0 +1,218 @@
...
+import 'package:stacked_tools/src/services/pubspec_service.dart';
+import 'package:stacked_tools/src/services/template_service_utils.dart';
+import 'package:stacked_tools/src/templates/compiled_template_map.dart';
+
+class RevertTemplateService {
+  final _fileService = locator<FileService>();
+  final _pubspecService = locator<PubspecService>();
+  final _clog = locator<ColorizedLogService>();
+  final _processService = locator<ProcessService>();
+  final _utils = locator<TemplateServiceUtils>();
```
FilledStacks: service variable name should match the service class name.

packages/stacked_tools/test/services_tests/render_template_service_test.dart
```diff
@@ -0,0 +1,267 @@
...
+import 'package:stacked_tools/src/services/render_template_service.dart';
+import 'package:stacked_tools/src/services/revert_template_service.dart';
+import 'package:stacked_tools/src/services/template_service_utils.dart';
+import 'package:stacked_tools/src/templates/compiled_template_map.dart';
+import 'package:stacked_tools/src/templates/template_constants.dart';
+import 'package:test/test.dart';
+
+import '../helpers/test_helpers.dart';
+
+RenderTemplateService _getRenderService() => RenderTemplateService();
```
FilledStacks: Common practice in tests is to call this `_getService()` it will be generated with the tests in the future. If you're testing a service you can get your service using `_getService()` if you're testing a ViewModel you can get it using `_getViewModel()`

packages/stacked_tools/test/services_tests/revert_template_service_test.dart
```diff
@@ -0,0 +1,855 @@
+import 'package:stacked_tools/src/locator.dart';
+import 'package:stacked_tools/src/services/revert_template_service.dart';
+import 'package:stacked_tools/src/templates/template_constants.dart';
+import 'package:test/test.dart';
+
+import '../helpers/test_helpers.dart';
+
+RevertTemplateService _getRevertService() => RevertTemplateService();
```
FilledStacks: Common practice in tests is to call this _getService() it will be generated with the tests in the future. If you're testing a service you can get your service using _getService() if you're testing a ViewModel you can get it using _getViewModel()

packages/stacked_tools/test/services_tests/template_service_test.dart
```diff
@@ -3,14 +3,18 @@ import 'package:stacked_tools/src/constants/message_constants.dart';
...
+import 'package:stacked_tools/src/services/template_service_utils.dart';
 import 'package:stacked_tools/src/templates/compiled_template_map.dart';
 import 'package:stacked_tools/src/templates/template_constants.dart';
 import 'package:test/test.dart';
 
 import '../helpers/test_helpers.dart';
 
-TemplateService _getService() => TemplateService();
+TemplateServiceUtils _getUtilsService() => TemplateServiceUtils();
+RenderTemplateService _getRenderService() => RenderTemplateService();
```
FilledStacks: This should be _getService

packages/stacked_tools/test/services_tests/template_service_test.dart
```diff
@@ -3,14 +3,18 @@ import 'package:stacked_tools/src/constants/message_constants.dart';
...
+import 'package:stacked_tools/src/services/revert_template_service.dart';
+import 'package:stacked_tools/src/services/template_service_utils.dart';
 import 'package:stacked_tools/src/templates/compiled_template_map.dart';
 import 'package:stacked_tools/src/templates/template_constants.dart';
 import 'package:test/test.dart';
 
 import '../helpers/test_helpers.dart';
 
-TemplateService _getService() => TemplateService();
+TemplateServiceUtils _getUtilsService() => TemplateServiceUtils();
```
FilledStacks: We shouldn't have multiple service's test in the same file. These need to be different testing files.

packages/stacked_tools/test/services_tests/template_service_test.dart
```diff
@@ -3,14 +3,18 @@ import 'package:stacked_tools/src/constants/message_constants.dart';
...
+RenderTemplateService _getRenderService() => RenderTemplateService();
+RevertTemplateService _getRevertService() => RevertTemplateService();
```
FilledStacks: This is not being used and should be removed.

FilledStacks (changes requested):
Lots of additional changes outside of the functionality being implemented in the PR so there's more rounds of reviews to come. 

Tip for the future. 

Make a PR for the feature you intend to build. Make another PR specifically for refactors. This PR will probably take a while to be merged in because so many parts of the code has changed, even if it's just naming.

PR #638, opened 2022-06-02
Feature/dialog generator

eb-so:
- Add dialog generator feature
- Add unit tests and integration tests


FilledStacks (approved):
Man, after the tests has been added it's so much easier to review this code. 

Happy with the implementation. Only one thing I'm not certain about and that's the file extension. .dialog or .dialogs (plural). I like plural more for this.

PR #646, opened 2022-06-06
Fixes Issues recently introduced on navigaiton

Eimen2018:
transition was causing error on replace with, clear stack and show etc


FilledStacks (approved):
LGTM.

This should have been caught if we navigated the old way. But now it covers both the new way and with extensions.

PR #651, opened 2022-06-08
Fix/routes is not required in stackedapp

### LINKED ISSUE #649: [Generator] Routes property is not required in the annotation but fails if not supplied
When you try to generate with stacked generator and you don't supply 'routes' property to the app setup, it throws an error when generating.

### **Error**
```
FormatException: Not an instance of list.
```
### **Error Screenshot**
Image from iOS

The workout for now is to pass an empty list to the route parameter
```
@StackedApp(
routes: [],
)
class App {}
```

### Expected
If `routes` list is required
- when it is not supplied to the `StackedApp` annotation, it should display a "route list is needed"

Or a default empty const list can be passed to the parameter.

eb-so:
Fix #649 
- When routes are not provided don't throw an exception 
- When routes are not provided don't generate anything

### DIFF
--- a/packages/stacked_generator/CHANGELOG.md
+++ b/packages/stacked_generator/CHANGELOG.md
@@ -1,3 +1,5 @@
+## 0.7.4
+- Prevent throwing an error when routes is not provided in StackedApp
 ## 0.7.3
 - Fixes Import for Adaptive Route
 ## 0.7.2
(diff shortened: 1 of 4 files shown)


Eimen2018 (approved):
LGTM

FilledStacks (approved):
LGTM

PR #652, opened 2022-06-08
test: :art: improves app navigation to include all navigation on integration tests

Eimen2018:


FilledStacks (approved):
LGTM. This was s good way to solve for all calls being used. Good move 🎉

PR #658, opened 2022-06-09
fix: :bug: strongly typed Id parameter never allowing view with id param

Eimen2018:


packages/stacked_generator/test/router/route_class_generator_test.dart
```diff
@@ -34,7 +34,7 @@ void main() {
     /// Note: It usually adds an extra line in the bottom of the result
     void generateRouteAndExpectResult(
         List<RouteConfig> routes, dynamic expectedResult,
-        {bool verbose = false}) {
```
eb-so: Why did you add this and not use it?

Eimen2018: It's leftover from the stash i use for generating the right constant 😁. Will remove it now

eb-so:
LGTM

PR #660, opened 2022-06-09
Fix/melos update

Eimen2018:
- Adds test command that runs build_runner and test on all app (package and example)


FilledStacks (approved):
LGTM! Can't wait to test it out.

FilledStacks (approved):
Awesome! It worked. We'll add the integration tests in there as well.

PR #663, opened 2022-06-10
Fix/future and stream viewmodels seterror

### LINKED ISSUE #486: StreamViewModel setError() not working.
I'm trying to use setError() in my view model (StreamViewModel), but it isn't reflected in dataReady or hasError.
I think this is because the parent class overrides getting of the error, but not setting.

Getting -
```dart
class _SingleDataSourceViewModel<T> extends DynamicSourceViewModel {
  dynamic _error;
  @override
  dynamic error([Object? object]) => _error;
}
```

Setting errors still uses BaseViewModel code -
```dart
class BaseViewModel extends ChangeNotifier {
  /// Sets the error for the ViewModel
  void setError(dynamic error) {
    setErrorForObject(this, error);
  }
  /// Sets the error state for the object equal to the value passed in and notifies Listeners
  /// If you're using a primitive type the value SHOULD NOT BE CHANGED, since Hashcode uses == value
  void setErrorForObject(Object object, dynamic value) {
    _errorStates[object.hashCode] = value;
    notifyListeners();
  }
}
```

So when accessing hasError and dataReady, the error set with setError is never acknowledged -
```dart
  // BaseViewModel
  bool get hasError => error(this) != null;
  // _SingleDataSourceViewModel<T>
  bool get dataReady => _data != null && !hasError;
```

Suggested fix, with commented possibilities based on initialise()'s streamSubscription.onError -
```dart
abstract class StreamViewModel<T> extends _SingleDataSourceViewModel<T>
    implements DynamicSourceViewModel, Initialisable {
  @override
  void setError(dynamic error) {
    // setErrorForObject(this, error); // Maybe? I'm not sure of relevance of _errorStates in _SingleDataSourceViewModels
    _error = error;
    // _data = null;
    // onError(error);
    notifyListeners();
  }
}
```

Thanks for all the helpful tutorials and stacked packages!

eb-so:
Closes #486 
  - Fix setError manually for futureviewmodel
  - Fix setError manually for streamviewmodel that last until the next emit of the stream

### DIFF
--- a/packages/stacked/CHANGELOG.md
+++ b/packages/stacked/CHANGELOG.md
@@ -1,3 +1,5 @@
+## 2.3.2
+- Fix [#486](https://github.com/FilledStacks/stacked/issues/486) 
 ## 2.3.11
 - Fix [#546](https://github.com/FilledStacks/stacked/issues/546) 
 ## 2.3.10
--- a/packages/stacked/lib/src/state_management/base_view_models.dart
+++ b/packages/stacked/lib/src/state_management/base_view_models.dart
@@ -226,7 +226,13 @@ class _SingleDataSourceViewModel<T> extends DynamicSourceViewModel {
 
   @override
   dynamic error([Object? object]) => _error;
+  @override
+  void setError(error) {
+    _error = error;
+    super.setError(error);
+  }
 
+  bool get hasError => error(this) != null;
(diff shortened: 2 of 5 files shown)


Eimen2018 (approved):
LGTM

FilledStacks (approved):
LGTM

PR #665, opened 2022-06-12
Fixes Mismatched Class Name and Constructor

### LINKED ISSUE #664: Mismatched Class Name and Constructor
When the generator is creating app.router.dart, there seems to be a scenario where the generator creates a class that has a mismatched class name and constructor.

For example:

```dart
/// ************************************************************************
/// Arguments holder classes
/// *************************************************************************

/// ExampleView arguments holder class
class NestedExampleViewArguments {
  final DateTime? date;
  final Key? key;
  ExampleViewArguments({this.date, this.key});
}
```

The scenario where this happens seems to be when a nested route defined in app.dart uses a page that has arguments. In this example, all other related generated code references ExampleViewArguments, and this class name is the only time NestedExampleViewArguments is referenced.

The expected outcome is that when navigating to the nested route and supplying arguments the arguments will be accepted. This was working in an earlier version.

In the latest version, the outcome is that the supplied arguments will be flagged as invalid and the app will not run.

Eimen2018:
- Fixes #664

### DIFF
--- a/packages/stacked/example/lib/app/app.router.dart
+++ b/packages/stacked/example/lib/app/app.router.dart
@@ -158,8 +158,14 @@ class BottomNavExampleRouter extends RouterBase {
   Map<Type, StackedRouteFactory> get pagesMap => _pagesMap;
   final _pagesMap = <Type, StackedRouteFactory>{
     FavoritesView: (data) {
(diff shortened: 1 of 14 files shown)


eb-so:
LGTM

FilledStacks (approved):
LGTM

PR #668, opened 2022-06-14
Feature/add custom text editing controller to form field

### LINKED ISSUE #384: [stacked_generator] [forms] Add support for passing custom controller for `FormTextField`
Sometimes, standard controllers are not enough.
It would be nice if we can specify a custom controller for the fields. 
I personally use a custom controller to add mask to the inputted text.(https://pub.dev/packages/extended_masked_text)
But right now I cant use this with the Stacked Form.

I will try to implement this and send a PR.

eb-so:
Fixes #384
- Split the base generator into interface and mixin
- Refactor form generator to builder pattern
- Feature - Add custom text editing controller to form field 
- Update stacked example


Eimen2018 (approved):
A lot of refactor...LGTM

eb-so:
> A lot of refactor...LGTM

Correct, also I think we should cover the form generator with more unit tests, cause we are going to add a lot of new features in the future

packages/stacked_generator/lib/src/generators/forms/stacked_form_content_generator.dart
```diff
@@ -1,14 +1,37 @@
...
+class StackedFormContentGenerator implements BaseGenerator {
   final FormViewConfig _formViewConfig;
-   late FormGeneratorUtil _util;
   StackedFormContentGenerator(this._formViewConfig);
 
+  @override
   String generate() {
-    _util = FormGeneratorUtil(formViewConfig: _formViewConfig);
-    return _util.generate();
+    final formGeneratorUtil = FormGeneratorUtil(formViewConfig: _formViewConfig)
```
FilledStacks: Given that this is now using the builder pattern I think this should be named `FormBuilder`

packages/stacked_generator/lib/src/generators/forms/stacked_form_content_generator.dart
```diff
@@ -1,14 +1,37 @@
...
+      ..writeHeaderComment()
+      ..generateImports()
+      ..generateValueMapKeys()
+      ..generateDropdownItemsMap()
+      ..generateTextEditingControllerItemsMap()
+      ..generateFocusNodeItemsMap()
+      ..generateValidationFunctionsFromAnnotation()
+      ..writeMixinSignature()
+      ..generateTextEditingControllersForTextFields()
```
FilledStacks: All `generate` commands should change to `add` to make sense in the domain language we're building in stacked.

FilledStacks (changes requested):
Some naming changes we have to apply to make things clearer in the code. 

This PR sets us up to confirm the order of calls for a builder as well, which will allow us to know if anything changes in order or output. 

There's a clear way to make this unit testable, it'll just take time and I don't know how to explain it in words haha.

PR #670, opened 2022-06-15
Feature/manual-form-validation

Eimen2018:
-  adds autoFieldValidation flag on formView
 -  adds validateForm on generated file
 -  adds isFormValid on extension

### DIFF
--- a/packages/stacked/example/lib/ui/form/example_form_view.form.dart
+++ b/packages/stacked/example/lib/ui/form/example_form_view.form.dart
@@ -85,7 +85,7 @@ mixin $ExampleFormView on StatelessWidget {
     shortBioController.addListener(() => _updateFormData(model));
   }
 
-  final bool autoTextFieldValidation = false;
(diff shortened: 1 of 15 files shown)


packages/stacked/example/lib/ui/form/example_form_view.form.dart
```diff
@@ -85,8 +85,14 @@ mixin $ExampleFormView on StatelessWidget {
     shortBioController.addListener(() => _updateFormData(model));
   }
 
+  final bool _autoTextFieldValidation = false;
+  bool validateFormFields(FormViewModel model) {
```
FilledStacks, line 89: This can be an extension on the viewmodel.

FilledStacks (approved):
LGTM

_Note_
I wanted to change the validate to an extension so we don't have to pass in `this` when calling it from inside the viewodel. But I don't think that's possible with our current inheritance structure.

PR #673, opened 2022-06-16
Ebrahim/fil 19 add bottomsheet setup code to generator

eb-so:
Add bottomsheet setup code to generator


packages/stacked_core/lib/src/code_generation/bottomsheets/stacked_bottomsheet.dart
```diff
@@ -0,0 +1,13 @@
+/// Described the Bottomsheet functionality to generate in the app
+class StackedBottomsheet {
```
Eimen2018, line 2: Can you please make it StackedBottomsheet<T> so that devs tell it what they expect to return i have added that on dialog to know what it expects to return on type safe

Eimen2018 (commented):
LGTM!

packages/stacked_core/lib/src/code_generation/bottomsheets/stacked_bottomsheet.dart
eb-so: What do I need to change exactly?

FilledStacks: @Eimen2018 type safety can be added in your PR. This scope has been covered here.

FilledStacks (approved):
LGTM

We'll tackle type safety after

PR #675, opened 2022-06-20
Feature/signin with phonenumber

Eimen2018:
- adds a method to sign in with phone number 
- adds a method to verifyOtp

### DIFF
--- a/packages/stacked_firebase_auth/lib/src/firebase_authentication_service.dart
+++ b/packages/stacked_firebase_auth/lib/src/firebase_authentication_service.dart
@@ -317,6 +317,23 @@ class FirebaseAuthenticationService {
     );
   }
 
+    /// Phone Number Login
+  Future<ConfirmationResult> signInWithPhoneNumber(String phoneNumber) async {
+    return firebaseAuth.signInWithPhoneNumber(phoneNumber);
+  }
+
+  Future<FirebaseAuthenticationResult> verifyOtp(
+      ConfirmationResult confirmationResult, String otp) async {
+    try {
+      UserCredential userCredential = await confirmationResult.confirm(otp);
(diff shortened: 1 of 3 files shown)


packages/stacked_firebase_auth/lib/src/firebase_authentication_service.dart
```diff
@@ -317,6 +317,23 @@ class FirebaseAuthenticationService {
     );
   }
 
+    /// Phone Number Login
+  Future<ConfirmationResult> signInWithPhoneNumber(String phoneNumber) async {
+    return firebaseAuth.signInWithPhoneNumber(phoneNumber);
```
eb-so: You should add a try/catch the same you did below

packages/stacked_firebase_auth/lib/src/firebase_authentication_service.dart
```diff
@@ -317,6 +317,23 @@ class FirebaseAuthenticationService {
...
+  }
+
+  Future<FirebaseAuthenticationResult> verifyOtp(
+      ConfirmationResult confirmationResult, String otp) async {
+    try {
+      UserCredential userCredential = await confirmationResult.confirm(otp);
+      return FirebaseAuthenticationResult(user: userCredential.user);
+    } catch (e) {}
+    return FirebaseAuthenticationResult.error(
```
eb-so: Place it inside the catch closer and add the error to the end of the error message

Eimen2018: @eb-so I thought i did 😆 ... my bad

eb-so:
LGTM

FilledStacks (approved):
Wooohooo! LGTM

PR #677, opened 2022-06-24
Add Sign In with phone number support for Native Platform

ferrarafer:
Adds Sign In with phone number support for Native Platform.

Also, `FirebaseAuthException` is exported to avoid firebase_auth dependency when using stacked_firebase_auth.

## Reference
The Firebase Authentication SDK for Flutter provides two individual ways to sign a user in with their phone number. Native (e.g. Android & iOS) platforms provide different functionality to validating a phone number than the web, therefore two methods exist for each platform exclusively:

Native Platform: verifyPhoneNumber.
Web Platform: signInWithPhoneNumber.

Reference: [link](https://firebase.google.com/docs/auth/flutter/phone-auth)

### DIFF
--- a/packages/stacked_firebase_auth/CHANGELOG.md
+++ b/packages/stacked_firebase_auth/CHANGELOG.md
@@ -1,6 +1,4 @@
 ## 0.2.13
-- Remove signInWithPhoneNumber (implements Web Platform)
-- Remove verifyOtp (Implements Web Platform)
 - Add requestVerificationCode (Implements Native Platform)
 - Add authenticateWithOtp (Implements Native Platform)
 
--- a/packages/stacked_firebase_auth/CHANGELOG.md
+++ b/packages/stacked_firebase_auth/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 0.2.13
+- Remove signInWithPhoneNumber (implements Web Platform)
+- Remove verifyOtp (Implements Web Platform)
+- Add requestVerificationCode (Implements Native Platform)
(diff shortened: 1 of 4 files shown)


eb-so:
LGTM, as long as it still works for the web

ferrarafer:
> LGTM, as long as it still works for the web

What do you mean? Are we using Stacked on web browsers? If that is the case, I can expose also Web Platform methods.

Eimen2018:
@ferrarafer We need the web too we are currently working on web 😂

ferrarafer:
> @ferrarafer We need the web too we are currently working on web 😂

@Eimen2018 Oki dokie, I will put back those methods for web in a few minutes.

eb-so:
@ferrarafer Beware web devs in the house

ferrarafer:
> @ferrarafer Beware web devs in the house

@eb-so haha my mistake, sorry. Web Platform support added back.

packages/stacked_firebase_auth/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 0.2.13
+- Remove signInWithPhoneNumber (implements Web Platform)
```
FilledStacks: This has not been removed? Can we update the changelog to match the latest hanges please

FilledStacks (changes requested):
We only need the change log updates to match what has been changed then this is ready to be merged.

FilledStacks (approved):

PR #679, opened 2022-06-28
Hide the `message` widget in `showCustomSnackBar`

eb-so:
- Hide the `message` widget in `showCustomSnackBar` when the message color is null and message is empty


FilledStacks (approved):
LGTM

PR #680, opened 2022-06-30
Fixes deeply nested imports & duplicate strongly typed methods

Eimen2018:
- adds a recognition to nested route imports 
- removes duplicated named route navigation method


eb-so:
LGTM

FilledStacks (approved):
LGTM

PR #689, opened 2022-07-05
Fix/logger not printing method name on flutter web

eb-so:
- Fix the logger not printing method name on flutter web
- Cleanup and refactor the logger code
- Remove deprecated compile check code


packages/stacked/example/lib/app/app.logger.dart
```diff
@@ -62,19 +61,50 @@ class SimpleLogPrinter extends LogPrinter {
...
+        final methodName = realFirstLine?.replaceAll('$className.', '');
+        return methodName;
+      }
     } catch (e) {
       // There's no deliberate function call from our code so we return null;
       return null;
     }
   }
+
+  List<String> _spliteClassNameWords(String className) {
```
FilledStacks: Is splite a word or should this be split?

FilledStacks (approved):
Just a single spelling mistake. The rest I'm happy with

packages/stacked/example/lib/app/app.logger.dart
eb-so: it's a typo

PR #692, opened 2022-07-07
Fix/dublicated types imports not added

eb-so:
- Fix dublicated types imports not added
- Update the stacked example
- Remove the rest of the guard code


packages/stacked_generator/CHANGELOG.md
```diff
@@ -1,8 +1,17 @@
 
-## 0.7.2
+## 0.7.13
+
+- Fixes dublicated types imports not added
```
Eimen2018, line 4: duplication* 😂

PR #697, opened 2022-07-22
Refactor/router generator to use aliased imports

### LINKED ISSUE #690: Stacked Generate doesn't alias clashing imports
## Problem
When you have packages that contain the models and both of those models area used in the view arguments the generated code is broken. 

In the project below I added google_maps_flutter and the map_toolkit package. Both of them have a `LatLng` model which breaks the generated code. 

Attached is a project that shows the bug. 

[Generator Bug Project](https://github.com/FilledStacks/stacked/files/9052190/generator_bug.zip)

## Solution
Add an alias to all imported packages in the router this will avoid all clashing of model type.

eb-so:
Fixes #690 
Refactor the router generator to use [CodeBuilder](https://github.com/dart-lang/code_builder)

@FilledStacks This is the output of the generator tested on your example
[example_fix.txt](https://github.com/FilledStacks/stacked/files/9180075/example_fix.txt)


FilledStacks (approved):
I really like this! What a great change. This will improve our code safety a lot. Only question now is how do we objectively prove it's always working. 

I see some unit tests, but this is such a big change I wouldn't want anyone losing time when we merge this in. 

@eb-so for the package version make it 0.8.0-beta.0 . This way it's not pushed to stable immediately and I can test it out in venu and you can test it on intake before we push it out to the world.

PR #707, opened 2022-07-28
Fixes assign import alias to view with a list parameter

eb-so:
- Fixes assign import alias to view with a list parameter


FilledStacks (approved):
LGTM! Thanks for quick turn around

PR #708, opened 2022-08-01
Call until() from Get instead Flutter's popUntil()

Hannnes1:
Previously, the `popUntil()` method in `NavigationService` called Flutter's method, instead of calling the one that comes with GetX. That meant that it did not work with nested navigators.

This pull request changes the method to call `until()` from GetX, but keeps the name in `NavigationService` as `popUntil()` to maintain backwards compatibility.

### DIFF
--- a/packages/stacked_services/CHANGELOG.md
+++ b/packages/stacked_services/CHANGELOG.md
@@ -1,3 +1,5 @@
+## 0.9.4
+- Adds `id` to the `popUntil` function in `NavigationService` to allow for nested navigation
 ## 0.9.3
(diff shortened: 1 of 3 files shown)


eb-so:
@Hannnes1 Nice addition 🙌
You just have to pump the stacked_services version number and add a message to the changelog so we can publish a new version

Hannnes1:
Done. I wasn't sure if you wanted to publish a new version directly, or do a bigger update later on.

eb-so:
@Hannnes1 Thanks for your contribution 🙌

PR #709, opened 2022-08-03
Feat/add instance name parameter to all dependency registration

eb-so:
- add `instanceName` parameter to all dependency registration constructers


Eimen2018 (approved):
LGTM - is this gonna be a breaking change ?

eb-so:
@Eimen2018 Nope

PR #712, opened 2022-08-06
refactor/stacked refactor

eb-so:
- Refactor `BaseViewModel` by extracting the different functionalities to mixins.
- Remove the inconsistency between the `BaseViewModel` and `FutureViewModel`, `StreamViewModel` when setting the error, data values.
- Now we have three mixins that can be mix with any viewmodel and they are:
  - `FutureRunnerHelper` that has `runBusyFuture` and `runErrorFuture` functions
  - `MessageStateHelper` that adds message proberty to the viewmodel
  - `DataStateHelper` that adds data proberty to the viewmodel
  - `FormStateHelper` that give the flexibility to use the `formviewmodel` with `futureviewmodel` for example
  - `IndexTrackingStateHelper` that give the flexibility to use the `IndexTrackingStateHelper` with `futureviewmodel` for example

Note: this change needs one small change in the form generator which is replace `FormViewModel` with `FormStateHelper`

### DIFF
--- a/packages/stacked/lib/src/state_management/helpers/index_tracking_state_helper.dart
+++ b/packages/stacked/lib/src/state_management/helpers/index_tracking_state_helper.dart
@@ -1,7 +1,6 @@
-import 'package:stacked/src/state_management/base_view_models.dart';
-import 'package:stacked/src/state_management/reactive_service_mixin.dart';
+import 'package:flutter/material.dart';
 
-class IndexTrackingViewModel extends ReactiveViewModel {
+mixin IndexTrackingStateHelper on ChangeNotifier {
   int _currentIndex = 0;
   int get currentIndex => _currentIndex;
 
@@ -22,7 +21,4 @@ class IndexTrackingViewModel extends ReactiveViewModel {
   }
 
   bool isIndexSelected(int index) => _currentIndex == index;
-
-  @override
-  List<ReactiveServiceMixin> get reactiveServices => [];
 }
--- a/packages/stacked/CHANGELOG.md
+++ b/packages/stacked/CHANGELOG.md
@@ -1,3 +1,11 @@
+## 2.4.0
+
+- Refactor `BaseViewModel` by extracting the different functionalities to mixins.
+- Remove the inconsistency between the `BaseViewModel` and `FutureViewModel`, `StreamViewModel` when setting the error, data values.
+- Now we have three mixins that can be mix with any viewmodel and they are:
+  - `FutureRunnerHelper` that has `runBusyFuture` and `runErrorFuture` functions
+  - `MessageStateHelper` that adds message proberty to the viewmodel
+  - `DataStateHelper` that adds data proberty to the viewmodel
 ## 2.3.15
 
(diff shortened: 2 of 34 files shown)


FilledStacks (approved):
This is great. I like the idea, but now there are going to be things we have to re-teach. 

I'd like the idea of a "Future functionality", "Stream Functionality" and "Reactive Functionality" to still exist. 

The refactor is good, but currently won't be easy to teach or be usable for the reason of the refactor. 

As long as everything developer facing stays the same I'm happy 🚀

packages/stacked/lib/src/state_management/helpers/index_tracking_state_helper.dart
```diff
@@ -1,7 +1,6 @@
-import 'package:stacked/src/state_management/base_view_models.dart';
-import 'package:stacked/src/state_management/reactive_service_mixin.dart';
+import 'package:flutter/material.dart';
 
-class IndexTrackingViewModel extends ReactiveViewModel {
+mixin IndexTrackingStateHelper on ChangeNotifier {
```
FilledStacks, line 3: Instead of this breaking change we can keep the viewmodels and make use the new mixin. This way no one's code will break when they updated but we can then write a tutorial or a blog post about the composable functionalites and explain how it can be used.

FilledStacks (changes requested):
The big request is to not remove old viewmodels until we've properly allowed users to learn about the composable functionality and choose if they want to use that. For instance, the views that are using IndexTrackingViewModel I don't want to update if I move up a version. But when I want to use multiple sets of functionalities like Future and Index then I can use the composable mixins. 

When we can avoid breaking changes we should. We have to think about the impact this would have on teams that need to release next week and suddenly has to go through and update all their code with nothing in the docs that tells them how to do it. 

So the changes requested in this review is:
- Ensure we still have all previous types available and working the same way
- The new types should be completely empty and use the new mixin functionality (This allows us to see if it works as expected before telling everyone to start using it)
- This should be a release that goes into the 0.8.0 beta so that we can test it out in venu first before it's released into the world. 

It'll allow me to catch all breaking changes before it goes out so that we can ensure we're making a stable release.

FilledStacks:
@eb-so am I right in saying that this is all in the beta branch and will be merged when master is merged? Would like to close this.

eb-so:
Yes you are correct

VittorioParagallo:
Autogenerated form code doesn't work for FormStateHelper.

In my use case I have a ViewModel like: 
`class MemberDetailsViewModel extends MultipleStreamViewModel with FormStateHelper{}`

And In the View:
  `@override
  void onViewModelReady(MemberDetailsViewModel viewModel) => syncFormWithViewModel(viewModel);`

But **syncFormWithViewModel autogenerated in the form file expects FormViewModel**. 
So to make it work i manually change the types in the autogenerated "member_details_view.form.dart"  (FormViewModel -> FormStateHelper)

I think is related to the first  @eb-so wrote in chronology
> Note: this change needs one small change in the form generator which is replace FormViewModel with FormStateHelper.

I just wanted to ask if it is a problem and there is a fix planned or I'm suppose to find another way?

may be i should open an issue?
Thanks a lot

PR #714, opened 2022-08-09
fixed stacked-cli not generating locator and router file

Edamijueda:
Running 'stacked create app <app_name>' in terminal using the CLI tool . Locator and Router file was not generated for the app folder in the new app. This is because the stacked_generator version in the stacked_tool package is old. This fix added the latest version of the stacked_generator.

### DIFF
--- a/packages/stacked_tools/lib/src/templates/app/pubspec.yaml.stk
+++ b/packages/stacked_tools/lib/src/templates/app/pubspec.yaml.stk
@@ -50,7 +50,7 @@ dev_dependencies:
   flutter_lints: ^1.0.0
   build_runner: ^2.2.0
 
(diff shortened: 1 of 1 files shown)


Eimen2018:
@Edamijueda there is a depreciated 'element' that you are using please change it to 'element2' that's why the unit test it failing or merge the latest master that might resolve it.

Eimen2018:
@Edamijueda this would be merged if you just pull master and push. also add package maintenance increase the version and update change log

Edamijueda:
I have merge, so my cloned version is currently up to date to filledStacks master. I guess I will have to create another pull request?
or is it possible to use this same issue to resolve?

Issue #717, opened 2022-08-13
[stacked_services] [snackbar] The SnackbarConfig.textColor not working as intended


dvagala:
I believe that the `SnackbarConfig.textColor` should define the color for all the text in a snackbar, unless otherwise specified explicitly in `titleColor`, `messageColor`, `mainButtonTextColor`.

**Steps to reproduce**
```
final snackbarService = locator<SnackbarService>();

snackbarService.registerSnackbarConfig(SnackbarConfig(
  textColor: Colors.red,
));

snackbarService.showSnackbar(
  message: 'The message',
  title: 'The title',
);
```

**Actual results**
Snackbar's title and message are not red.

**Expected results**
Snackbar's title and message are red.

Eimen2018:
Thank you for the PR

PR #718, opened 2022-08-13
Fix/snackbar text color

dvagala:
Fix for #717 issue + some releated cleanup

### DIFF
--- a/packages/stacked_services/lib/src/snackbar/snackbar_service.dart
+++ b/packages/stacked_services/lib/src/snackbar/snackbar_service.dart
@@ -73,7 +73,7 @@ class SnackbarService {
               title,
               key: Key('snackbar_text_title'),
               style: TextStyle(
-                color: _snackbarConfig?.titleColor,
+                color: _snackbarConfig?.titleColor ?? _snackbarConfig?.textColor ?? Colors.white,
                 fontWeight: FontWeight.w800,
                 fontSize: 16,
               ),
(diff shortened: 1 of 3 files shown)


packages/stacked_services/lib/src/snackbar/snackbar_service.dart
```diff
@@ -68,7 +68,7 @@ class SnackbarService {
     Get.snackbar(
       title,
       message,
-      titleText: _snackbarConfig?.titleColor != null || title.isNotEmpty
```
dvagala, line 71: I don't see a reason why a `_snackbarConfig?.titleColor` should play a role whether the title widget is visible, so I deleted those conditions. Corrent me if I'm missing something

packages/stacked_services/lib/src/snackbar/snackbar_service.dart
```diff
@@ -68,31 +68,30 @@ class SnackbarService {
...
               style: TextStyle(
-                color: _snackbarConfig?.messageColor,
+                color: _snackbarConfig?.messageColor ?? _snackbarConfig?.textColor ?? Colors.white,
                 fontWeight: FontWeight.w300,
                 fontSize: 14,
               ),
               textAlign: _snackbarConfig?.messageTextAlign ?? TextAlign.left,
             )
           : SizedBox.shrink(),
-      colorText: _snackbarConfig?.textColor ?? Colors.white,
```
dvagala, line 95: This field is not used because we are always setting the title and message style widget explicitly

Eimen2018:
@dvagala you also need to pull master here too 😅

Issue #724, opened 2022-08-16
Supported platforms are not specified


dvagala:
Devs cannot see what platforms are supported for this package on [pub.dev](https://pub.dev/packages/stacked). There is a [UNKNOWN PLATFORMS]. This may be reason for someone not using the package for their project.

dvagala:
Oh, it's an [issue](https://github.com/dart-lang/pana/issues/1113) on the pub.dev side. I'm closing this

PR #725, opened 2022-08-16
add supported platforms explicitly

dvagala:
See the issue #724 

In the [docs](https://dart.dev/tools/pub/pubspec#platforms) they say to add the supported platforms explicitly in pubspec.yaml , but never done this before, so please correct me if I'm wrong.

I believe the `stacked` package is supporting all the platforms, there's no native code or dependency. Personally tested iOS, Android, web, macOs


dvagala:
Oh, it's an [issue](https://github.com/dart-lang/pana/issues/1113) on the pub.dev side. I'm closing this

PR #726, opened 2022-08-22
Fix/line-length-delete-service

Eimen2018:
- Adds formatting only for updated files
- Adds formatting line length for all commands
- Adds delete service command


eb-so (commented):
LGTM

FilledStacks (approved):
LGTM

PR #727, opened 2022-08-24
Feat : Support for Duration in SnackbarConfig

vishhmakasana:
Duration field added in snackbar_config.dart so that we will not have to specify Duration on each show snackbar call.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Stacked [![Pub Version](https://img.shields.io/pub/v/stacked)](https://pub.dev/packages/stacked)
 
-[![Stacked 3.1 release banner](https://github.com/FilledStacks/stacked/blob/master/packages/stacked/assets/banner.jpeg)](https://stacked.filledstacks.com/)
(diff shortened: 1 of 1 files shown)


packages/stacked_services/lib/src/snackbar/snackbar_service.dart
```diff
@@ -160,6 +160,9 @@ class SnackbarService {
               textAlign: snackbarConfig.titleTextAlign,
             )
           : snackbarConfig.titleText ?? null,
+      /// Please check if its intentionally or some typo mistake : it should be
+      // messageText: snackbarConfig.messageColor != null || message.isNotEmpty
```
Eimen2018: Yes i think it should be snackbarConfig.messageColor

Eimen2018 (changes requested):
LGTM, please update the comment and add package maintenance (increase version and update change log)

FilledStacks (approved):
Thanks for the PR and Thanks @Eimen2018 for the review.

We have an automatic deploy and update of changelog and version now 😄

FilledStacks (changes requested):
Only request is to fix merge issues and then we can move forward

PR #729, opened 2022-08-31
Update main.dart

1shaked:
The accent color is deprecated so we can not use it; we have to use color scheme


1shaked:
fix the example

PR #737, opened 2022-10-04
Suggestion: setting the state with 'setState'

dvagala:
Hey, I've always had my own extensions on BaseViewModel, and I'm putting one of them here as a suggestion to integrate into the package. Nothing crucial but maybe also other people can benefit from that.

Suggestion:
I never liked the fact that I always need to call the `notifyListeners()` after setting the state in ViewModel. The suggestion is to use a `setState()` callback. It encapsulates the whole concept of notifying UI, and does exactly what it should - **Setting the State**. It's subjective, but for me, the readability is much better.

```dart
class AnyViewModel extends BaseViewModel {
  int _postCount = 0;
  int get postCount => _postCount;

  // Old Approach
  void setPostCount(int count) {
    _postCount = count;
    notifyListeners();
  }

  // New Approach
  void setPostCount(int count) {
    setState(() {
      _postCount = count;
    });
  }
}

```

### DIFF
--- a/packages/stacked/lib/src/state_management/base_view_models.dart
+++ b/packages/stacked/lib/src/state_management/base_view_models.dart
@@ -23,6 +23,16 @@ class BaseViewModel extends ChangeNotifier
     disposed = true;
     super.dispose();
   }
+
+  void setState(void Function()? fn) {
+    fn?.call();
+    notifyListeners();
+  }
+
+  Future<void> setStateAsync(Future<void> Function()? fn) async {
+    await fn?.call();
+    notifyListeners();
+  }
 }
 
 /// A [BaseViewModel] that provides functionality to subscribe to a reactive service.


FilledStacks:
Hey, that's a good suggestion. In this simple situation it might look better to some, it's a lot of extra code for the same functionality. 

If you have a function that notiifies listeners at specific points and conditionally then this falls apart. 

```dart

Future<void> createAccount() async {

// Do some work to get an account verification token from Plaid

notifyListeners(); // This will update to the new state in the UI


// Process the account link results

notifyListeners();

... etc

}

```

It would be A LOT more code with setState. I think it's a good extension to have but I don't think it should be apart of stacked. It will also most likely cause confusion when I talk about not using UI code in the viewmodel because it's so close to how the Flutter framework does state management.

dvagala:
Yes yes I understand. It was just a suggestion. I'm closing the PR

PR #750, opened 2022-10-28
Added finer error inspection

JohnnyRainbow81:
You can check the viewModel for hasAnyError() and currentError() now.

### DIFF
--- a/packages/stacked/example/lib/main.dart
+++ b/packages/stacked/example/lib/main.dart
@@ -22,6 +22,6 @@ class MyApp extends StatelessWidget {
       theme: ThemeData(
         primarySwatch: Colors.blue,
       ),
-    );
+    ); 
   }
 }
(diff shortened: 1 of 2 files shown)


FilledStacks (changes requested):
Awesome. 

You need to do the general housekeeping as well. 

- Update the patch version number
- Add an entry into the changelog for the new version number

JohnnyRainbow81:
Heya...
@FilledStacks  could you please share some info on how to choose a patch version number? 
Cause this absolutely is my first time contributing to open source...

eb-so:
@JohnnyRainbow81 Stacked versioning policy
For versioning a good guide is.

- When we collect a number of minor improvements and features we bump the major version
- When we implement new functionality we increase the minor version (second number)
- If we fix something or add to an existing implementation .i.e. add a new transition we increase the patch number (last digit)
- If no code changes or we only update configuration and the functionality of everything is still the same we add a +version

JohnnyRainbow81:
Thank you @eb-so ! I'll investigate tonight, update the PR and then we can hopefully proceed.

PR #753, opened 2022-11-10
Fix/name property in router generator not reflected on navigator extensions

### LINKED ISSUE #740: [stacked_generator] Not import both classes with same name from two different libraries
Hi,
i have two classes with same name, but under different paths and libraries of the same package.
When i importing both in file app.dart, then i set keyword (import with AS) to one of them (or for both of them, the result is same) and after success generate procedure, file app.router.dart contain only one imported class of them.
Please, what i make a wrong?

eb-so:
Fixes #740


FilledStacks (approved):
Great @eb-so I'll merge this and release with the other fixes

PR #756, opened 2022-11-10
Add listeners count getter

AdamH4:
Sometimes it is useful when reactive service has knowledge about its listeners. You can then make decisions base on that. For example I would like to cancel some listener but if there are listeners I won't do cancellation.


okursoftware (commented):
LGTM

FilledStacks:
Looks good. 

Could you please bump the minor version and add an entry into the changelog with your changes.

FilledStacks (commented):
I will add this to the class, we've renamed some file. 

So I'll close this, but it will be added 😄

PR #757, opened 2022-11-12
Fix/missing prefix of enum default value in router generator

### LINKED ISSUE #745: [stacked_generator] Missing prefix of enum default value in app.router.dart
I have a view with default value is an enum in constructor like this:

```dart
class CameraView extends StatelessWidget {
  final List<CameraMode> cameraModes;
  final bool allowSwitchCamera;
  const CameraView({
    super.key,
    this.cameraModes = const [
      CameraMode.image,
      CameraMode.video,
    ],
    this.allowSwitchCamera = false,
  });
}
```
app.router.dart
image

eb-so:
Fixes #745 

- Exclude the default values imports from being aliased by adding a custom Allocator that Exempt all the import that ends with #EXCLUDE word

### DIFF
--- a/packages/stacked/example/lib/app/app.router.dart
+++ b/packages/stacked/example/lib/app/app.router.dart
@@ -6,9 +6,10 @@
 
 // ignore_for_file: no_leading_underscores_for_library_prefixes
 import 'package:example/app/custom_route_transition.dart' as _i7;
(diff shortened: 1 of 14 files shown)


FilledStacks (approved):
Awesome!

PR #768, opened 2022-11-17
Fix/type safe navigation generating the wrong casing

### LINKED ISSUE #766: Type safe navigation generating the wrong casing
If you add a route like below to your `StackedApp`

```dart
 CupertinoRoute(page: ChatView)
```

the type safe navigation extension generated is `navigateTochatView` and should be `navigateToChatView`

eb-so:
Fix #766


FilledStacks (approved):
Nice! Thanks @eb-so

PR #770, opened 2022-11-20
Improvements on stacked_tools

ferrarafer:
- Rename binary to stacked_tools.dart
- Fix testHelpersPath default value

> The test helpers filename is fixed at the moment, test_helpers.dart, we can only configure the path from the config.

- Change default values on stacked config template

> Set the default values on the stacked config to match the naming conventions.

- Add test_helpers filename to service_test template
- Load config on create app command

> Without this, the user was not able to create an app with stacked_tools with different configuration because default values were used.


ferrarafer:
@FilledStacks 
I would like to remove widget_test.dart from test folder when create app command is executed. Because, I think is not needed and also breaks the tests. I don't know the best way to do it yet.

packages/stacked_tools/lib/src/models/config_model.dart
```diff
@@ -24,7 +24,7 @@ class Config with _$Config {
     /// Relative path for test helper file that contains the register functions
     /// for unit test setup and mock declarations.
     @JsonKey(name: 'test_helpers_path')
-    @Default('test/helpers/test_helpers.dart')
+    @Default('test/helpers')
```
FilledStacks, line 27: Adding a new test file does some weird stuff with the imports. I wonder if this will fix it.

FilledStacks (changes requested):
Hi, we have to change the binary back to stacked.

If we move the package to stacked_cli on pub.dev then we can change the binary to stacked_cli. But I don't like stacked_tools in the command.

packages/stacked_tools/lib/src/models/config_model.dart
ferrarafer: I think with this changes would work fine.

FilledStacks (changes requested):
@ferrarafer I think a smart thing for us to do is to release this as beta so that I can test it out in venu first. 

Could you update the version to 0.2.3-beta.0 then I'll merge and we can release and test it out without affecting anyone that might be using the cli.

FilledStacks (approved):

PR #772, opened 2022-11-21
Fix typed navigation on nested views

### LINKED ISSUE #767: Typesafe navigation should use name provided in Routes
Additional bug:

When giving a name to a nested navigation route, the generated type safe (?) navigation uses the pagename instead of the passed name parameter.

```dart
  MaterialRoute(page: NestedNavigatorView, name: "nestedRouter", children: [
    MaterialRoute(page: HomeView, initial: true),
  ]), 
```

The above generates:

```dart
Future<dynamic> navigateToNestedHomeViewInNestedNavigatorView([
    ...
  ]) async {
    return navigateTo<dynamic>(NestedNavigatorViewRoutes.homeView,
        id: routerId,
        preventDuplicates: preventDuplicates,
        parameters: parameters,
        transition: transition);
  }
```

Notice how `navigateTo` uses `NestedNavigatorViewRoutes` instead of `NestedRouterRoutes`.

ferrarafer:
When route.className.key is different of route.name, we should use name because has higher priority, otherwise, we use route.className.key.

Fixes #767

### DIFF
--- a/packages/stacked/CHANGELOG.md
+++ b/packages/stacked/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 3.0.0+2
+
+- Fix [#767](https://github.com/FilledStacks/stacked/issues/767)
+
 ## 3.0.0+1
 
 - Replace the old `ReactiveValues` approach with `notifyListeners()` when using the `ReactiveServiceMixin` in readme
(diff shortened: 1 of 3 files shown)


FilledStacks (changes requested):
Add the version and change log updates and this is good to go!

FilledStacks (approved):
@ferrarafer great. 

In terms of versioning (we should probably write this in the contribution guide).

- When we do a fix on existing functionality we bump the patch (last digit)
- When we add minor new functionality we bump the minor version (second digit)
- When we have collected enough of a change set or bigger features we bump major version (first digit)
- When there's a documentation change or something in the readme with no code changes we use the + at the end. 

I'll approve this and merge and bump the patch version instead.

PR #777, opened 2022-11-25
Add onDispose method to ViewModelBuilderWidget

ferrarafer:
Add onDispose method to ViewModelBuilderWidget to match the same functionality as the old way method, probably the method was omitted by accident.


FilledStacks (approved):
Awesome!

PR #778, opened 2022-11-25
add listen param to getParentViewModel function

Dionnie:
Added listen param so we can tell Provider to not rebuild or rebuild the widget.


FilledStacks (approved):
Thanks @Dionnie123

PR #771, opened 2022-11-20
Stacked improvements

ferrarafer:
- Added ListenableServiceMixin
- Deprecated ReactiveServiceMixin
- Use ListenableServiceMixin instead of ReactiveServiceMixin
- Change reactiveServices property name on ReactiveViewModels to listenableServices
- Move data models to data_models folder
- Move SkeletonLoader to ui folder
- Rename state_management folder to view_models
- Fix unit test file names to follow conventions

PR #781, opened 2022-11-29
[cli] Add support for `v1` view style

ferrarafer:
- Changes View template to use StackedView, aka ViewModelWidget, as default
- Adds config option `v1` to use alternative view builder style
- Adds flag option `v1` to use alternative view builder style, overrides config option
- Adds `v1`support for create app command
- Removes `widget_test` after on create app command
- Uses relativeLocatorPath on every template
- Fixes TestHelpers import on views and services tests
- Improves getFilePathToHelpersAndMocks
- Updates example app to align changes


ferrarafer:
I made two PRs with the same solution but different implementation on TemplateService.

Probably we can find another solution more generic, this is kind of ad hoc.

packages/stacked_tools/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.2.4-beta.0
```
FilledStacks, line 1: For beta we keep the version the same and use beta.1 instead. So next version stays 0.2.3, we only update the beta.X number

packages/stacked_tools/lib/src/models/config_model.dart
```diff
@@ -46,6 +46,14 @@ class Config with _$Config {
     @JsonKey(name: 'register_mocks_function')
     @Default('registerServices')
         String registerMocksFunction,
+
+    /// Boolean value to determine view builder style
+    ///
+    /// This is used when creating a view file during `create view` command. By
+    /// default, ViewModelWidget style is used.
+    @JsonKey(name: 'use_view_model_builder_style')
```
FilledStacks: I think we should shorten this. I prefer something like

```shell
stacked create view login --v1
```

or to make it clearer

```shell
stacked create view login --useBuilder
```

@ferrarafer this actually reminds me that we can make the name better for the new style widget. I think we should call it `StackedView<T>` or ``ViewModelView<T>` I like both. One builds our brand, the other is aligned with MVVM. 

Let me know what you think

packages/stacked_tools/lib/src/services/config_service.dart
```diff
@@ -64,6 +64,12 @@ class ConfigService {
         _customConfig.testViewsPath,
       );
 
+  /// Returns boolean value to determine view builder style.
+  ///
+  /// False: ViewModelWidget
+  /// True: ViewModelBuilder
+  bool get useViewModelBuilderStyle => _customConfig.useViewModelBuilderStyle;
```
FilledStacks: Oooooooh. I'm only seeing now. This is in the stacked_config. Good choice! We should be able to override as well from the command.

FilledStacks (changes requested):
Some changes requested, but I was a bit confused here. So lets chat through the comments.

packages/stacked_tools/CHANGELOG.md
ferrarafer: I changed the patch version because it adds a new configuration option, but probably could be part of the previous version. What leaves me with doubts is we are also changing the default view template.

packages/stacked_tools/lib/src/services/config_service.dart
ferrarafer: I will add the override through command.

packages/stacked_tools/CHANGELOG.md
FilledStacks: @ferrarafer you're right. 

We can leave it like this for now.

packages/stacked_tools/lib/src/models/config_model.dart
ferrarafer: I like `StackedView` 👍

ferrarafer:
@FilledStacks 
PR #771 needs to be approved and merged before merging because on the last updates I migrate to StackedView.

Otherwise, let me know and I can update this PR to continue using ViewModelWidget.

FilledStacks:
@ferrarafer it's been merged. You can complete this PR and request a review when you're done.

packages/stacked_tools/lib/src/commands/create/create_app_command.dart
```diff
@@ -20,21 +25,34 @@ class CreateAppCommand extends Command {
...
   Future<void> run() async {
-    await locator<ConfigService>().loadConfig();
+    await _configService.loadConfig();
 
     final appName = argResults!.rest.first;
 
     await _processService.runCreateApp(appName: appName);
 
+    /// Removes `widget_test` file to avoid failing unit tests on created app
+    await _fileService.deleteFile(filePath: '$appName/test/widget_test.dart');
```
FilledStacks, line 46: Interesting. 

We should add this to the template actions and add a delete action. This way we can provide a "step" in the template that deletes a file as well. 

This solution works for now. If we need to do another deletion we can change it into a delete action in the template, which I think we have already.

FilledStacks (approved):
Awesome work @ferrarafer 

I left a note on the delete action. For now it's okay, but that should be apart of the create app template. 

.i.e. One of the modifications should be to delete the widget_test.dart file.  That can be done later. I'm happy with this.

PR #782, opened 2022-11-29
View builder style [Implementation 2]

ferrarafer:
- Change View template to use ViewModelWidget style as default
- Adds config option to set view builder style

Closes #781

### DIFF
--- a/packages/stacked_tools/lib/src/templates/compiled_templates.dart
+++ b/packages/stacked_tools/lib/src/templates/compiled_templates.dart
@@ -151,12 +151,15 @@ const String kAppTemplateStackedConfigJsonStkPath =
 
 const String kAppTemplateStackedConfigJsonStkContent = '''
 {
-    "views_path": "lib/ui/views",
-    "services_path": "lib/services",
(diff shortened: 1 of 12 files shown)


packages/stacked_tools/lib/src/templates/compiled_templates.dart
```diff
@@ -1088,6 +1091,40 @@ class {{viewName}} extends StatelessWidget {
...
+const String kViewTemplateGenericViewPath =
+    'lib/ui/views/generic/generic_view.dart.stk';
+
+const String kViewTemplateGenericViewContent = '''
+import 'package:flutter/material.dart';
+import 'package:stacked/stacked.dart';
+
+import '{{viewModelFileName}}';
+
+class {{viewName}} extends ViewModelBuilderWidget<{{viewModelName}}> {
```
FilledStacks, line 1105: Refer to the previous PR's review. 

Lets think about a better name for this so it's not too confusing. It should be easy to remember either based on the package we're using or the architecture we're using.

FilledStacks (changes requested):

PR #784, opened 2022-12-01
[stacked] Fix version and improve changelog

ferrarafer:


packages/stacked/CHANGELOG.md
```diff
@@ -1,16 +1,19 @@
-## 3.0.1-beta.0
+## 3.0.2-beta.0 [BREAKING CHANGES]
```
FilledStacks: With breaking changes we should update he minor version. 

Lets make this 3.1.0-beta.0 . All changes required for us to write the docs better and improve will be added into 3.1.0 and that will be our release announcement version. So you can keep adding to 3.1.0 going forward and only update the beta.X number.

For the migration guide below there's a few grammar issues. Use the following copy in place. 

"In our mission to make Stacked a standalone framework we want to formalise some ideas. This showed us that some of our widgets are not named appropriately. In addition to that there are also places where we use viewModel and others where we use model.  To fix that we made the following breaking changes all within the `ViewModelBuilderWidget`"

FilledStacks (changes requested):
Some copy updates and a guide to stick on 3.1.0-beta.X until we release the docs with migration into the world. It all has to be ready when we release otherwise we might break a lot of projects 😅

FilledStacks (approved):

PR #786, opened 2022-12-14
Add line length support to Stacked config

ferrarafer:
- Add lineLength to Config model
- Add line-length option support to create app command
- Fix replaceCustomPaths, now only process values which keys contains `path` word


FilledStacks (approved):
Great PR @ferrarafer 🔥 Merging and releasing

PR #787, opened 2022-12-16
Fix Stacked Hooks

ferrarafer:
- Replaced HookViewModelWidget with StackedHook
- Updated example app
- Updated README


FilledStacks (approved):
Merging this and then I'll do some updates to change the `StackedHookView` since `StackedHook` sounds like a hook that we wrote.

PR #788, opened 2022-12-16
Update builder to short hand

ferrarafer:
- Updated viewModelBuilder on templates to use short hand function notation
- Updated ViewModel instance name from model to viewModel
- Updated StackedView template builder indentation


FilledStacks (changes requested):
Awesome. I want us to rely on the CI again so I'll leave this open until we fix the CI. I might attempt to do that on Sunday as well if you don't get to it.

PR #794, opened 2022-12-19
Improve stacked_tools

ferrarafer:
- Removed lib and test folder from the default paths because there is no need to include them as they can't be changed
- Changed getImportPath to sanitizePath, more appropriate method name
- Removed path of app name, if any, to render template


FilledStacks (changes requested):
See comments from latest PR to apply to this. We can probably close this one and apply in the newest PR? @ferrarafer

ferrarafer:
We are going to merge #795 instead of this which also include this modifications.

PR #795, opened 2022-12-21
Add XDG_CONFIG_HOME support

ferrarafer:
- Changed Stacked config filename
- Changed ConfigService to support XDG_CONFIG_HOME
- Moved again loadConfig to each command instead of bootstrap
- Moved `xdg_directories` dependency to PathService
- Updated `Config` model default values
- Added warning message for users when using deprecated Stacked config filename
- Added warning message for users when using deprecated paths on Stacked config file
- Added sanitize function for config file to remove unnecessary `lib` or `test` parts on paths
- Fix unit tests

This PR was made on top of PR #794 so if it is approved, #794 should be closed.


packages/stacked_tools/lib/src/constants/config_constants.dart
```diff
@@ -1,2 +1,2 @@
-/// Defines the path of the stacked configuration file
-const String kConfigFilePath = 'stacked.config.json';
+/// Defines the filename of the stacked configuration file
+const String kConfigFileName = 'stacked.json';
```
FilledStacks, line 2: I like this change, but I think if we're changing it now, and there's people already using it we should make it backwards compatible. 

We don't have to modify a lot of the code to get that working. What we want to do is:
1. Check for the new config file name and perform everything as expected
2. If it's not there, then check for deprecated config file name `stacked.config.json`. When we find this, print a noticeable log that says. "Stacked config file should be renamed from stacked.config.json to stacked.json. Stacked cli will not read stacked.config.json files after the next minor release."
3. Use the stacked.config.json if it's there and continue running the code as normal.

packages/stacked_tools/CHANGELOG.md
```diff
@@ -1,4 +1,14 @@
+## 0.2.4-beta.4
+
+- Removed lib and test folder from the default paths because there is no need to include them as they can't be changed
```
FilledStacks, line 3: For this change, I want us to provide a warning to the users that has older configs so that they can update their file. SImilar to the comment on the path change that I left. 

We want to have this functionality in the code now so we can know how to make it scalable as we most likely will change other parts of the cli that we want to notify the user about in the future.

FilledStacks (changes requested):
Awesome, some small user warning changes that I'd like us to add in.

FilledStacks (approved):
Awesome! Gonna be trying this out soo

FilledStacks (approved):
Trying out after merge and release.

PR #805, opened 2023-01-03
Adds generate command

FilledStacks:
Adds `stacked generate` as a command which is an alias for `flutter pub run build_runner build --delete-conflicting-outputs`


ferrarafer:
LGTM ✅

PR #808, opened 2023-01-04
Add generate command

ferrarafer:
Adds `stacked generate` as a command which is an alias for `flutter pub run build_runner build --delete-conflicting-outputs`.

Replicate PR #805 made by @FilledStacks because I missed it on migration to `stacked_cli` 😅


FilledStacks (approved):
hahaha

PR #812, opened 2023-01-04
Deprecate stacked_tools package

ferrarafer:
- Deprecated `stacked_tools` package.
- Added error message about package been moved

@FilledStacks I show the error message first, and after that the commands run as always. Do you want to keep it like this or do you want to remove the functionality? I mean, do you want only to show the message and exit?


FilledStacks (changes requested):
@ferrarafer happy with the implementation. 

I think you should rebase from the latest master in stacked. There's some changes here that's getting reverted. 

Also, I think you should start doing PR's from within the repo using branches. It'll be easier to pull latest code from master 😄

ferrarafer:
@FilledStacks let me know if the rebase was as you wanted

FilledStacks:
@ferrarafer it's merged. we can release the latest version later today.

PR #814, opened 2023-01-06
Add analytics to stacked cli

### LINKED ISSUE #775: Stacked Cli Analytics
To ensure we know how many users are using our cli and which commands are most important. We need to add instrumentation around the usage of the cli. 

This task requires research to check what we can use (some kind of fire + forget functionality to an analytics api) where we can log the information required for cli usage.

ferrarafer:
We need to assign the correct tracking ID (UA-XXXXXXXX-X) to the AnalyticsService before merge.

### DIFF
--- a/packages/stacked_cli/lib/src/services/analytics_service.dart
+++ b/packages/stacked_cli/lib/src/services/analytics_service.dart
@@ -2,7 +2,7 @@ import 'package:usage/usage_io.dart';
 
 class AnalyticsService {
   final AnalyticsIO _analytics = AnalyticsIO(
-    'UA-XXXXXXXX-X',
+    'UA-41171112-5',
     'stacked-cli',
     'stacked_cli',
   );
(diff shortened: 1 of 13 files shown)


packages/stacked_cli/lib/src/services/analytics_service.dart
```diff
@@ -0,0 +1,62 @@
+import 'package:usage/usage_io.dart';
+
+class AnalyticsService {
+  final AnalyticsIO _analytics = AnalyticsIO(
+    'UA-XXXXXXXX-X',
```
FilledStacks: Can we move this into a secure place? Is that required? 

If someone malicious sees this they might spam us with analytics 😅

FilledStacks (changes requested):
Awesome. Lets figure out how we'll store the project ID. I'll create one and send it to you.

FilledStacks:
Lol, I don't know where to get this. We'l figure it out on the call later.

FilledStacks (approved):
Awesome

PR #815, opened 2023-01-06
Replace double quotes with simple quotes

ferrarafer:
This change was made only for consistency on quotes used. A user on Slack, `Niraj Nandish`, was having issues with their linting.


FilledStacks (approved):
Doooope!

PR #820, opened 2023-01-12
Replace stacked_tools strings with stacked_cli

ferrarafer:
There was some references to `stacked_tools` on the help output.


FilledStacks (approved):
Awesome!

PR #822, opened 2023-01-13
Add create bottom_sheet command

### LINKED ISSUE #783: Stacked Cli generate bottom sheet
Similar to the views we should be able to scaffold bottom sheets with viewmodels and without. This command should:
- Create the new enum value (if there isn't one then create the enum)
- Register the builder to the setupBottomSheetsUi function
- Create the sheet view file with completer and request passed in
- Create the viewmodel for the sheet (If the option is true, by default it should be true)
- Create the viewmodel tests for the sheet if above is true

ferrarafer:
- Added create bottom_sheet command
- Improved templates
- Improved analytics responsiveness, no more delays on commands execution (#827 needs to be merged first)
- Updated example app


packages/stacked_cli/example/lib/ui/setup/setup_bottom_sheet_ui.dart
```diff
@@ -1,18 +1,23 @@
 import 'package:flutter/material.dart';
 import 'package:example/app/app.locator.dart';
 import 'package:example/enums/bottom_sheet_type.dart';
-import 'package:example/ui/bottom_sheets/notice_sheet/notice_sheet.dart';
+import 'package:example/ui/bottom_sheets/notice/notice_sheet.dart';
+// @stacked-bottom-sheet-import
```
FilledStacks: We can use the existing template identifier for this `// @stacked-import`

packages/stacked_cli/lib/src/templates/app/lib/ui/setup/setup_bottom_sheet_ui.dart.stk
```diff
@@ -1,18 +1,23 @@
 import 'package:flutter/material.dart';
 import 'package:{{packageName}}/{{{relativeLocatorPath}}}';
 import 'package:{{packageName}}/enums/bottom_sheet_type.dart';
-import 'package:{{packageName}}/ui/bottom_sheets/notice_sheet/notice_sheet.dart';
+import 'package:{{packageName}}/ui/bottom_sheets/notice/notice_sheet.dart';
+// @stacked-bottom-sheet-import
```
FilledStacks: Use the `// @stacked-import` identifier here since we only have the 1 import.

packages/stacked_cli/lib/src/templates/app/lib/ui/setup/setup_bottom_sheet_ui.dart.stk
```diff
@@ -1,18 +1,23 @@
...
 import 'package:stacked_services/stacked_services.dart';
 
+typedef BottomSheetBuilder = Widget Function(
```
FilledStacks, line 8: Nice update

packages/stacked_cli/lib/src/templates/template_render_functions.dart
```diff
@@ -24,7 +24,7 @@ Map<String, RenderFunction> renderFunctions = {
   kTemplateNameService: (ReCase value, [map]) {
     final configService = locator<ConfigService>();
     return {
-      kTemplatePropertyServiceName: value.pascalCase,
+      kTemplatePropertyServiceName: '${value.pascalCase}Service',
```
FilledStacks, line 27: Awesome. Good to see how comfortable you are with the codebase.

FilledStacks (changes requested):
Very happy with this PR. Small change to use the existing `// @stacked-import` instead of creating a new import import identifier specifically for the bottom sheet. 

 

I didn't see a docs PR to add this functionality. I've updated [the task ](https://www.notion.so/filledstacks/Add-create-bottom-sheet-into-stacked-cli-235f883ea47c46cc8931762539fb376d) to include the additions to be made to the docs when this functionality is merged.

packages/stacked_cli/example/lib/ui/setup/setup_bottom_sheet_ui.dart
ferrarafer: You are right, I don't know why I didn't think of that yesterday.

packages/stacked_cli/lib/src/templates/app/lib/ui/setup/setup_bottom_sheet_ui.dart.stk
ferrarafer: I was happy with the change and I known u would like it 😄

packages/stacked_cli/lib/src/services/analytics_service.dart
```diff
@@ -1,66 +1,97 @@
...
     'UA-41171112-5',
     'stacked-cli',
     'stacked_cli',
   );
 
-  Future<void> appCreated({required String name}) async {
+  /// This will wait until all outstanding analytics requests have completed,
+  /// or until the specified duration has elapsed.
+  Future<void> waitLastPingOrCloseAtTimeout() async {
+    await _analytics.waitForLastPing(
```
FilledStacks, line 14: Good catch. I didn't know about this. I hope this doesn't mean that we could potentially lose some analytics.

FilledStacks (approved):

PR #823, opened 2023-01-13
Remove unnecessary code

ferrarafer:
@FilledStacks I believe that file is not used, what do you think?


FilledStacks (approved):
I agree. I think it's an old file

PR #827, opened 2023-01-14
Improve analytics

ferrarafer:
- Removed compile command event from analytics
- Moved analytic calls to top of the command to trigger the event first and avoid delay
- Used waitForLastPing to avoid delay on analytic events

### DIFF
--- a/packages/stacked_cli/lib/src/services/analytics_service.dart
+++ b/packages/stacked_cli/lib/src/services/analytics_service.dart
@@ -7,7 +7,7 @@ class AnalyticsService {
     'stacked_cli',
   );
 
-  Future<void> appCreated({required String name}) async {
+  Future<void> createAppEvent({required String name}) async {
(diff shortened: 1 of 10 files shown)


packages/stacked_cli/lib/src/services/analytics_service.dart
```diff
@@ -1,75 +1,97 @@
...
   final AnalyticsIO _analytics = AnalyticsIO(
     'UA-41171112-5',
     'stacked-cli',
     'stacked_cli',
   );
 
-  Future<void> appCreated({required String name}) async {
+  /// This will wait until all outstanding analytics requests have completed,
+  /// or until the specified duration has elapsed.
+  Future<void> waitLastPingOrCloseAtTimeout() async {
```
FilledStacks, line 13: I never knew about this. good find @ferrarafer

FilledStacks (approved):

PR #828, opened 2023-01-15
Fix support for configurable paths

ferrarafer:
- Added support for bottom sheet type enum file path on Stacked config
- Added support for bottom sheet builder file path on Stacked config
- Added support for append a template on special cases
- Added support for relative router file path (Fixes an old bug we didn't realize was there)


FilledStacks (approved):
LGTM

FilledStacks:
@ferrarafer lets get the tests passing again and then I'll merge this in.

PR #834, opened 2023-01-19
Track errors with analytics on stacked cli

ferrarafer:
Logged exception events to Google Analytics.

**NOTE**: I logged some warning level events, let me know if you want to remove them and only keep error level events.


FilledStacks (approved):
Good choice on adding the warnings as well. Now we'll get an accurate refelection of the users experiences.

PR #835, opened 2023-01-20
Add global options for enable or disable sending analytics data and show version number

### LINKED ISSUE #775: Stacked Cli Analytics

### LINKED ISSUE #839: [stacked_cli] Respect Opt-out/Opt-in for Google Analytics
Hey!

Since you are using the `usage`-Package for Google Analytics you should follow the Opt-out/Opt-in convention like stated in the documentation:

### When do we send analytics data?

```
You can use this library in an opt-in manner or an opt-out one. It defaults to opt-out - 
data will be sent to Google Analytics unless the user explicitly opts-out. 
The mode can be adjusted by changing the value of the Analytics.analyticsOpt field.

Opt-out In opt-out mode, if the user does not explicitly opt-out 
of collecting analytics, the usage library will send usage data.

Opt-in In opt-in mode, no data will be sent until the user explicitly opt-in to collection. 
This includes screen views, events, timing information, and exceptions.
```

I think it is ok to use it as an opt-out but i would like to see a setting, maybe in your config file to deactivate it. I would not use the cli in a company enviroment atm.

ferrarafer:
- Added global option to show stacked_cli version
- Added global option to enable sending of analytics data
- Added global option to disable sending of analytics data
- Added prompt for sending analytics at first run


FilledStacks:
Once you change what we discussed I will review, merge and release.

ynnob:
I would like to see a way to disable it regardless of the enviroment. See #839

packages/stacked_cli/bin/stacked.dart
```diff
@@ -1,30 +1,118 @@
...
+
+  if (argResults[ksDisableAnalytics]) {
+    locator<AnalyticsService>().enable(false);
+    return true;
+  }
+
+  return false;
+}
+
+Future<void> _handleFirstRun() async {
```
FilledStacks, line 98: Yooohoooooo! This is 🔥  @ferrarafer

FilledStacks (approved):
Hahaaaa, awesome update!

PR #836, opened 2023-01-20
best practice on showing form validation message

rsenthilkumar6:
If would be best to skip if condition on view and use visibility widget instead. new comers & learners referring the example would make use of it as a habit


FilledStacks (approved):
@rsenthilkumar6 much appreciated!  I'll merge this in and fix integration tests separately.

PR #843, opened 2023-01-24
Fix version global option

ferrarafer:
@FilledStacks I don't know if this can be added to `stacked_generator` to avoid running generate command on every version bump. For now, fixes the issue. I don't know if it is worth it to improve it.


FilledStacks (approved):
We can leave it like this for now, but I think it would work great in the compile command.

PR #846, opened 2023-01-26
Refactor ProcessService and version global option

ferrarafer:
- Added `_runProcess` private function which replaces `_runProcessAndLogOut` and `_runProcessInSilence` functions
- Added `runPubGlobalList` to retrieve a list of dart packages installed
- Refactored `runAnalyze` function to avoid using a temporal file
- Refactored `_handleVersion` to get version on the fly


FilledStacks (approved):
Nice implemetation.

FilledStacks:
Merging even with the CI failure. I see it's only analyzer issues which we can fix separately

PR #848, opened 2023-01-30
[stacked_cli] Add model flag option to bottom sheets and dialogs

ferrarafer:
- Added `model` flag option to create bottom_sheet command to use or not a model
- Added `model` flag option to create dialog command to use or not a model


FilledStacks (approved):
Aewsome!

PR #849, opened 2023-01-31
[generator] Align BottomSheets and Dialogs generators with Stacked

ferrarafer:
- Changed builder name from dialog to dialogs
- Removed `Sheet` word at end of enum value on BottomSheets
- Removed `Dialog` word at end of enum value on Dialogs
- Added Stacked template identifiers
- Changed Map type for builders, more precise

Merges after #848 because it fixed GitHub Actions.

### DIFF
--- a/packages/stacked_generator/lib/src/generators/bottomsheets/generate/bottomsheet_class_generator_helper.dart
+++ b/packages/stacked_generator/lib/src/generators/bottomsheets/generate/bottomsheet_class_generator_helper.dart
@@ -9,7 +9,6 @@ class BottomsheetClassGeneratorHelper with StringBufferUtils {
     for (var bottomsheetConfig in bottomsheetConfigs) {
       writeLine('${bottomsheetConfig.enumValue},');
     }
-    writeLine('// @stacked-bottom-sheet-type');
     writeLine("}");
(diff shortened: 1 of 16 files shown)


packages/stacked_generator/lib/src/generators/bottomsheets/generate/bottomsheet_class_generator_helper.dart
```diff
@@ -1,14 +1,13 @@
...
-import 'package:stacked_generator/src/generators/extensions/string_utils_extension.dart';
+import 'package:stacked_generator/src/generators/bottomsheets/bottomsheet_config.dart';
 
-import '../bottomsheet_config.dart';
 import 'bottomsheet_class_content.dart';
 
 class BottomsheetClassGeneratorHelper with StringBufferUtils {
   void writeBottomsheetTypeEnum(List<BottomsheetConfig> bottomsheetConfigs) {
-    writeLine("enum BottomsheetType{");
+    writeLine("enum BottomSheetType{");
```
FilledStacks, line 8: Good catch

FilledStacks (approved):
Awesome. Will release before the end of my day.

PR #861, opened 2023-02-15
feat: return parameter values on Argument class for better accessibility

### LINKED ISSUE #831: [stacked_generator] Add toString and toMap to the route arguments class
Currently if someone tries to access the route arguments they're shown `Instance of <Route>Arguments`(Route being the class name of the view) which doesn't help if you need the arguments for Google Analytics or any other similar service. For example, if you have a view named `NewsTutorialView` with some arguments. When you go to that view and pass in the required arguments for it, the `route.settings.arguments` holds a value of `Instance of NewsTutorialViewArguments`. 

It would be better if the arguments class builder added a custom `toString` and `toMap` function for each argument class it generates for `app.router.dart`

Sembauke:
closes #831


packages/stacked_generator/lib/src/generators/router/generator/arguments_class/arguments_class_builder.dart
```diff
@@ -10,16 +10,31 @@ class ArgumentsClassBuilder {
...
       return Class(
         (b) => b
           ..name = argumentsBuilderHelper.argumentClassName
           ..fields.addAll(argumentsBuilderHelper.convertParametersToClassFields)
-          ..constructors.add(argumentsBuilderHelper.argumentConstructer),
+          ..constructors.add(argumentsBuilderHelper.argumentConstructer)
+          ..methods.add(
+            Method(
+              (b) => b
+                ..name = 'getParametersAsString'
```
FilledStacks: I think this should override the `toString()` instead

packages/stacked_generator/lib/src/generators/router/generator/arguments_class/argument_class_builder_helper.dart
```diff
@@ -26,6 +26,14 @@ class ArgumentClassBuilderHelper {
         .toList();
   }
 
+  List<String> get convertParametersToStrings {
```
FilledStacks: Can we add a duplicate of this to convert the parameters to a Map. I think that would be quite useful to print out in that way. Key = name, value  = parameterValue.

packages/stacked_generator/lib/src/generators/router/generator/arguments_class/arguments_class_builder.dart
```diff
@@ -10,16 +10,31 @@ class ArgumentsClassBuilder {
...
+      final parametersAsString =
+          argumentsBuilderHelper.convertParametersToStrings.toString();
+
       return Class(
         (b) => b
           ..name = argumentsBuilderHelper.argumentClassName
           ..fields.addAll(argumentsBuilderHelper.convertParametersToClassFields)
-          ..constructors.add(argumentsBuilderHelper.argumentConstructer),
+          ..constructors.add(argumentsBuilderHelper.argumentConstructer)
+          ..methods.add(
```
FilledStacks: Add another function called toMap() that allows you to return the arguments as a map.

FilledStacks (changes requested):
Thank you for the contribution. I'm happy with this implementation. I think the naming can be a bit better and I'd also like to see how this looks in the example app. 

See comments below for naming. 

If you go to the `stacked` package under the packages directory, there's an example app in there. If you can run the generator in there, it'll always choose the local version of the package so it'll generate the new code so we can inspect and see how that looks.

FilledStacks:
Also, I just realised it's not ready for review, but I had some time so you can tackle those changes while it's still in Draft mode 😄

Sembauke:
Yes, I was planning on adding the map function too. 

Also, I think the best thing we can do is override the toString and toMap functionality. As the argument value carries an instance of (Class)Arguments, which we would add this behaviour to.

I was not convinced just putting the values in to a List would be the best way to go. I am not sure entirely what you suggested but In my mind it looked something like this:

`"key=value,key=value"`

image


for the map functionality it would look like:

`{'key': 'value', 'key': 'value'}`

I've tested these changes against my own application, but testing it here would make it easier.

Sembauke:
I just noticed that there is no toMap function bound on Object. We also can't seem to override any other methods.

Sembauke:
@FilledStacks 
what do you think about adding this on the navigationService instead? It seems to be in the same scope and we won't have to worry about which functionality we can override?

something like this:
```
_navigationService.getRouteArgumentsAsString(Route.(name))
```

Nirajn2311:
> I just noticed that there is no toMap function bound on Object. We also can't seem to override any other methods.

Yes that's correct so the idea I had was to first create the `toMap` function and then override the `toString` function to print the map as a string

FilledStacks:
@Sembauke I think the map print out looks fine. It's common for developers to read pairings like that, without the `=`. I didn't even know there's a toMap override 😅 I was thinking of the same implementation that @Nirajn2311 mentions. Create a toMap, then override toString to use that function so it's always printed in a structured way.

Sembauke:
Alright, we can do that. I guess I misunderstood what you wanted.

It should be good now:
image

Sembauke:
The Argument Class is an instance of an Object, which can be any type of Object according to Dart. 

So you can't add extra functions to the argument class without overriding. It is somehow not possible add a null check for this also. 

So I decided to override the toString with a string map.

FilledStacks:
@Sembauke nice, I'm a bit confused by the `required this.` that's a part of the key. I think we should only have the parameter name as the key in the map. 

Also, can you bump the minor version + add the update into the Changelog. 

And when you're ready for this to be reviewed you can take it out of draft mode and I'll do another full review on it.

Sembauke:
I thought it might be useful to add this, and required. But now I think about it, I can't think of any use case were that might be useful.

I assumed the version on `stacked_generator` should have been updated. Let me know what you think of the change log.

Sembauke:
@FilledStacks, could you run the CI and see if all the test pass?

FilledStacks (approved):

FilledStacks (changes requested):
Happy with the updates. 

The only thing I'd like to see is how it looks when generated as apart of the PR. 

Please run `stacked generate` in the `packages/stacked/example` folder. If you don't have stacked_cli installed you can run `flutter pub run build_runner build --delete-conflicting-outputs`

Then commit the changes so we have it as apart of the PR.

FilledStacks (approved):
Doooooope! Merging and then I will release right after

FilledStacks:
@Sembauke Could rebase from our master to resolve the branch conflicts please. I pushed a generator update so there's a clash in the versioning I'm guessing in the changelog. 

### Rebase Instructions

git remote add upstream https://github.com/FilledStacks/stacked.git
git fetch upstream
git pull --rebase upstream/master master `OR (if syntax is not working)` git pull --rebase upstream master
git push origin master --force

Sembauke:
@FilledStacks Should be good now hopefully..

FilledStacks:
To avoid more back and forth I'm going to merge this in instead of rebase. Then I'll release it. 

Thank you for all your effort.

Sembauke:
Yeah, no problem. Sorry for the back and forth. But it is in finally 😄

Nirajn2311:
@FilledStacks Thanks for the help

PR #862, opened 2023-02-17
Two Way Text Field Binding

### LINKED ISSUE #382: Set form values from view model.
Hi @FilledStacks, great work with the FormView decorator :D.

I've a question:

Right now It's possible to retrieve data from forms in the view model. But I've not see any way to set the form data from the view model?

For example: one app with a logic to retrieve data from the server to allow the user to edit the form data.

How can I do that?

I see in the generated `*.form.dart` does have in the extension the `get` operation, but does not have the `set` operation.

FilledStacks:
Adds the ability to set the value of the TextEditting controller by calling `[fieldName]Value = 'Text';`

The Values in the viewmodel are now bound to the TextEdittingController values for the form field. 

This makes setting initial values or values from remote sources in the text field much simpler. 

Fixes #382

### DIFF
--- a/packages/stacked_generator/lib/src/generators/forms/form_builder.dart
+++ b/packages/stacked_generator/lib/src/generators/forms/form_builder.dart
@@ -405,6 +405,18 @@ class FormBuilder with StringBufferUtils {
           'String? get ${caseName.camelCase}ValidationMessage => this.fieldsValidationMessages[${_getFormKeyName(caseName)}];');
     }
 
+    // Write out the clearForm method
+    writeLine('void clearForm() {');
+    for (final field in fields) {
+      if (field is! TextFieldConfig) {
+        continue;
+      }
+
+      final caseName = ReCase(field.name);
+      writeLine("${caseName.camelCase}Value = ' '; ");
+    }
+    writeLine('}');
+
     writeLine('}');
     return this;
   }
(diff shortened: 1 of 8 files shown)


packages/stacked_generator/lib/src/generators/forms/form_builder.dart
```diff
@@ -405,6 +405,18 @@ class FormBuilder with StringBufferUtils {
           'String? get ${caseName.camelCase}ValidationMessage => this.fieldsValidationMessages[${_getFormKeyName(caseName)}];');
     }
 
+    // Write out the clearForm method
+    writeLine('void clearForm() {');
```
ferrarafer, line 409: I would add a newLine() above this line.

`newLine();`

ferrarafer (approved):
LGTM 🔥 

I only added a comment of a newLine for readability.

FilledStacks:
Merging in. I'll fix the CI in the PR I'm about to make for the new template functionality

chrisvidal:
brilliant feature !
how about the `FormDateField` ? we stil have to set them trough the [fieldName]Map ?

FickleLife:
@FilledStacks any chance you might be able to explain how I can set values on the form from the model using this method? I am not sure where I call `[fieldName]Value = 'Text';`?

PR #867, opened 2023-02-19
[cli] avoid version check on compile and update command

ferrarafer:
- Avoids running pub command if has last version
- Avoids notify new version available for compile and update commands


FilledStacks (approved):
Good catch. I saw this happening as I was using the compile command.

PR #872, opened 2023-02-22
Fixing extended ChangenNotifier

Sipioteo:
When you have a class that exenteds ChangeNotifier, and a child who extends the parent, the function wasn't able to recognize it. With this little change, it will be able to recognize the ChangeNotifier everywhere in the extension tree.


FilledStacks:
Hey @Sipioteo Thank you for the PR! 

Can you please update the patch number of the version and add it to the changelog.

Issue #874, opened 2023-02-23
[Feature] Ability to supply org and platform


felixkriechbaum:
Hey folks,

I think I have a useful idea for the cli.
It would be cool if you could also specify an --org and a --platform like in flutter create

for example:
```shell
stacked create app demo --org com.test --platform ios,android
```

would be really cool

chetanchoudhary08:
would be very helpful to have the above where user can put values for package name and language of android whether kotlin or java. much needed

FilledStacks:
Interesting. I like that idea. I'll tag it. We won't get to this soon, but you're welcome to make a PR if you're comfortable.

yangricardo:
Not saw this before, but also opened this
#974

ferrarafer:
Feature added to the new CLI release.

PR #893, opened 2023-03-11
Add the posibility to add the platform and an organization

felixkriechbaum:
Hello all,

regarding the following FR which I mentioned the other day (#874) I thought I'd try my hand at it and created something that does what I mentioned.

There are 2 new "arguments" that you can now specify when creating a stacked app.

This would be -o / --org followed by the bundle identifier
```shell
-o com.filledstack 
```
or
```shell
--org com.filledstack
```

and the same for the platform but with -p --platform followed by the platforms separated by ,
 
```shell
-p windows,linux 
```
or
```shell
--platform windows,linux
 ```

Also, I have a kind of check of the app name according to the rules of dart (The name should be written in lowercase, with underscores to separate the words, just_like_this. Use only basic Latin letters and Arabic numerals: [a-z0-9_]. Also make sure that the name is a valid dart identifier, i.e. does not start with digits and is not a reserved word) and that the bundle identfier needs a '.' (com.example)

I have tested it as far as possible, but of course strange or untested things can always happen

Please be merciful with me, it's my first PR :D
For questions or improvement wishes I am gladly open

### DIFF
--- a/packages/stacked_cli/lib/src/commands/create/create_app_command.dart
+++ b/packages/stacked_cli/lib/src/commands/create/create_app_command.dart
@@ -35,6 +35,18 @@ class CreateAppCommand extends Command {
     );
 
     argParser.addOption(
+      "org",
+      abbr: "o",
+      help: kCommandOrganizationHelp,
+    );
+
+    argParser.addOption(
+      "platform",
+      abbr: "p",
+      help: kCommandPlatformHelp,
+    );
+
+    argParser.addOption(
       ksLineLength,
       abbr: 'l',
       help: kCommandHelpLineLength,
@@ -55,15 +67,62 @@ class CreateAppCommand extends Command {
   @override
   Future<void> run() async {
     await _configService.loadConfig();
+    _cLog.stackedOutput(message: argResults!.rest.first, isBold: true);
     final appName = argResults!.rest.first;
     final appNameWithoutPath = appName.split('/').last;
     final templateType = argResults![ksTemplateType];
+    String organization = "";
+    String platforms = "";
+
+    int indexOrg = argResults!.arguments.indexOf("--org");
+    if (indexOrg != -1) {
+      organization = argResults!.arguments[indexOrg + 1];
+    }
+
+    if (indexOrg == -1) {
+      indexOrg = argResults!.arguments.indexOf("-o");
+      if (indexOrg != -1) {
+        organization = argResults!.arguments[indexOrg + 1];
+      }
+    }
+
+    int indexPlatform = argResults!.arguments.indexOf("--platforms");
+    if (indexPlatform != -1) {
+      platforms = argResults!.arguments[indexPlatform + 1];
+    }
+
+    if (indexPlatform == -1) {
+      indexPlatform = argResults!.arguments.indexOf("-p");
+      if (indexPlatform != -1) {
+        platforms = argResults!.arguments[indexPlatform + 1];
+      }
+    }
+
+    // Organization must be a valid bundle name. (com.example)
+    if (organization.isNotEmpty && !organization.contains(".")) {
+      _cLog.error(
+          message: "Organization must be a valid bundle name. (com.example)");
+      return;
+    }
+
+    // Regarding https://dart.dev/tools/pub/pubspec#name
+    // The name may only contain lowercase letters, digits, underscores.
+    // And is not allowed to starts with numbers.
+    if (appName.contains(RegExp(r'[A-Z]')) ||
+        appName.startsWith(RegExp(r'[0-9]'))) {
+      _cLog.error(
+          message:
+              "App name must contains lower letters and is not allowed to start with numbers (example, example1)");
+      return;
+    }
 
     unawaited(_analyticsService.createAppEvent(name: appNameWithoutPath));
     _processService.formattingLineLength = argResults![ksLineLength];
-    await _processService.runCreateApp(appName: appName);
 
     _cLog.stackedOutput(message: 'Add Stacked Magic ... ', isBold: true);
+
+    await _processService.runCreateApp(
+        organization: organization, platforms: platforms, appName: appName);
 
     await _templateService.renderTemplate(
       templateName: name,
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: packages/stacked_cli/lib/src/commands/create/create_app_command.dart [hunk 1]
```diff
@@ -55,16 +67,63 @@ class CreateAppCommand extends Command {
   @override
   Future<void> run() async {
     await _configService.loadConfig();
+    _cLog.stackedOutput(message: argResults!.rest.first, isBold: true);
```
FilledStacks, line 70: Please remove this log. It looks like it was used for debugging.

packages/stacked_cli/lib/src/constants/command_constants.dart
```diff
@@ -3,6 +3,8 @@
 const String ksDart = 'dart';
 const String ksFlutter = 'flutter';
 const String ksCreate = 'create';
+const String ksOrganization = "--org";
```
FilledStacks, line 6: This is the only constant used in the command line that has -- in front of it. Please follow the convention of the rest of the option implementations. Look at template for reference.

packages/stacked_cli/lib/src/commands/create/create_app_command.dart
```diff
@@ -55,16 +67,63 @@ class CreateAppCommand extends Command {
...
+    if (indexOrg != -1) {
+      organization = argResults!.arguments[indexOrg + 1];
+    }
+
+    if (indexOrg == -1) {
+      indexOrg = argResults!.arguments.indexOf("-o");
+      if (indexOrg != -1) {
+        organization = argResults!.arguments[indexOrg + 1];
+      }
+    }
```
FilledStacks, line 87: - The code has to be DRY (Don't Repeat Code). You have multiple lines of the exact same code in a row, 4 times in a row. 
- None of the other commands use this convention you're using to check if an option was supplied. Please make sure you understand how the other commands to it and use the same way. 
- No Magic values in our code unless there's an explicit instruction that it has to be generated in the future from the compile command. Please remove all magic values
- You don't have to check the option for both versions, again, look at the implementation of the other options. Follow along with the one that closes resemble what you want to implement and copy the structure and implementation of that.

packages/stacked_cli/lib/src/commands/create/create_app_command.dart
```diff
@@ -55,16 +67,63 @@ class CreateAppCommand extends Command {
...
+        platforms = argResults!.arguments[indexPlatform + 1];
+      }
+    }
+
+    // Organization must be a valid bundle name. (com.example)
+    if (organization.isNotEmpty && !organization.contains(".")) {
+      _cLog.error(
+          message: "Organization must be a valid bundle name. (com.example)");
+      return;
+    }
```
FilledStacks, line 106: This check can be refactored out of here and unit tested on its own.

packages/stacked_cli/lib/src/commands/create/create_app_command.dart
```diff
@@ -55,16 +67,63 @@ class CreateAppCommand extends Command {
...
+    // Regarding https://dart.dev/tools/pub/pubspec#name
+    // The name may only contain lowercase letters, digits, underscores.
+    // And is not allowed to starts with numbers.
+    if (appName.contains(RegExp(r'[A-Z]')) ||
+        appName.startsWith(RegExp(r'[0-9]'))) {
+      _cLog.error(
+          message:
+              "App name must contains lower letters and is not allowed to start with numbers (example, example1)");
+      return;
+    }
```
FilledStacks, line 117: This check should be refactored out of there and unit tested

packages/stacked_cli/lib/src/services/process_service.dart
```diff
@@ -27,10 +27,26 @@ class ProcessService {
...
+    List<String> arguments = [];
+    arguments.add(ksCreate);
+
+    if (organization.isNotEmpty) {
+      arguments.addAll([ksOrganization, organization]);
+    }
+
+    if (platforms.isNotEmpty) {
+      arguments.addAll([ksPlatform, platforms]);
+    }
```
FilledStacks, line 43: This pattern is causing a lot of additional code that's not required. you can simply make the value null and use null to know if a value was added or not. Making it required and passing empty is not a good pattern.

packages/stacked_cli/lib/src/services/process_service.dart
```diff
@@ -27,10 +27,26 @@ class ProcessService {
...
+
+    arguments.add(appName);
```
FilledStacks, line 45: This can all be done in line in the array definition using the latest dart syntax. 

```dart
[
 if(..)
  value,
value,
]
```

packages/stacked_cli/lib/src/commands/create/create_app_command.dart
```diff
@@ -34,6 +34,18 @@ class CreateAppCommand extends Command {
       help: kCommandHelpV1,
     );
 
+    argParser.addOption(
+      "org",
```
FilledStacks, line 38: Please remove these string literals and follow the pattern of the other options.

FilledStacks (changes requested):
Hi, thanks for the PR. 

1. You should test your code before you make changes and after you've made changes (this doesn't work)
2. If you don't know how to test your code you should spend all your time figuring that out first before you write the code. Submitting code that doesn't work hands over responsibility you took on to the code reviewer
3. If your code doesn't work your work is not done
4. Follow coding principles that I teach in my videos + the basic software engineering principles like keeping code DRY
5. When you see conventions in a codebase try and follow those. The thing you want to do, unless a completely new feature, would have something similar in the code you can follow. In this case there are many commands that take options. You should follow those commands implementations. 

The biggest take away is that you don't know if your code works (It doesn't), and you should rather spend time figuring out to test it because there's a lot you didn't implement here. 

- Templates are not updated and compiled into the cli
- Template rendering is not updated (writing your value into the template)

felixkriechbaum:
Will take a look and reopen when finished

PR #914, opened 2023-03-23
Update Readme.md to reflect different URLs

SpeerJ:
It looks like some of the repositories got moved around. I corrected the URLs.


FilledStacks (approved):
Thank you!

PR #937, opened 2023-04-28
Feature/add org and app name

sebastianbuechler:


FilledStacks:
Hi @sebastianbuechler sorry for not tending to this. The PR was also not complete as the templates weren't updated. There's quite a bit of work around the CLI to add something new. This is something that I'll be fixing soon.

PR #943, opened 2023-05-11
removed vsync from skeleton_loader

masnun-siam:


ferrarafer:
@masnun-siam we already merged #941 but we really appreciate the collaboration, thanks !

Issue #918, opened 2023-03-26
[bug]: Error: No named parameter with the name 'vsync'.


Al-Krause-ex:
### Errors:
_pub.dev/stacked-3.2.0/lib/src/view_models/ui/skeleton_loader.dart:106:19: Error: No named parameter with the name 'vsync'.
                  vsync: this,
                  ^^^^^_
_lib/src/widgets/animated_size.dart:28:9: Context: Found this candidate, but the arguments don't match.
  const AnimatedSize({
        ^^^^^^^^^^^^_

### My code:
**lib/core/app/routes.dart**
```import 'package:stacked/stacked_annotations.dart';
import 'package:todos_frontend/presentation/show_todos/show_todos_view.dart';

@StackedApp(routes: [
 AdaptiveRoute(page: ShowTodosView, initial: true),
])
class App {}
```

**routes.router.dart**
```// GENERATED CODE - DO NOT MODIFY BY HAND

// **************************************************************************
// StackedNavigatorGenerator
// **************************************************************************

// ignore_for_file: no_leading_underscores_for_library_prefixes
import 'package:flutter/material.dart';
import 'package:stacked/stacked.dart' as _i1;
import 'package:stacked_services/stacked_services.dart' as _i3;
import 'package:todos_frontend/presentation/show_todos/show_todos_view.dart'
    as _i2;

class Routes {
  static const showTodosView = '/';

  static const all = <String>{showTodosView};
}

class StackedRouter extends _i1.RouterBase {
  final _routes = <_i1.RouteDef>[
    _i1.RouteDef(
      Routes.showTodosView,
      page: _i2.ShowTodosView,
    )
  ];

  final _pagesMap = <Type, _i1.StackedRouteFactory>{
    _i2.ShowTodosView: (data) {
      return _i1.buildAdaptivePageRoute<dynamic>(
        builder: (context) => const _i2.ShowTodosView(),
        settings: data,
      );
    }
  };

  @override
  List<_i1.RouteDef> get routes => _routes;
  @override
  Map<Type, _i1.StackedRouteFactory> get pagesMap => _pagesMap;
}

extension NavigatorStateExtension on _i3.NavigationService {
  Future<dynamic> navigateToShowTodosView([
    int? routerId,
    bool preventDuplicates = true,
    Map<String, String>? parameters,
    Widget Function(BuildContext, Animation<double>, Animation<double>, Widget)?
        transition,
  ]) async {
[...]

PR #944, opened 2023-05-11
Remove vsync property on AnimatedSize widget

ferrarafer:
Fixex #918


ferrarafer:
I missed #941 so I will close this.

PR #955, opened 2023-05-17
fixed : No named parameter with the name 'force'.

Imesh7:
https://github.com/Stacked-Org/stacked/issues/954


Imesh7:
this issue is occurs when using stacked : ^ 3.2.6
I am fixed issue with downgrade the plugin to stacked : 3.2.4
remove symbol of `^` before version number

`stacked : 3.2.4`

PR #963, opened 2023-05-31
feat: set currentIndex from route on Web Platform

### LINKED ISSUE #911: [bug]: bottom nav bar example shows wrong selected nav bar item after browser refresh
### Describe the bug

The bottom nav bar example demonstrates the usage of the `NestedRouter()` with an `AppBar()` Widget. Selecting the NavBar Item 'History' e.g. will select the item and leads to this url: `.../bottom-nav-example/history-view`. After refreshing the browser, the url is still the same as expected and also shows the `HistoryView` but the favorites icon is selected. 

### To reproduce

1. Run router_example
2. Navigate to history view via the app bar item
3. Refresh the browser
4. HistoryView is shown but Favorites Icon is selected

### Expected behavior

Since there are many scenarios where Nested Routing is used with an appbar the expected behavior in web would be to see the correct nav bar item selected. 

### Screenshots

Bildschirmfoto 2023-03-22 um 14 18 22

### Additional Context

_No response_

ferrarafer:
Closes #911

### DIFF
--- a/lib/src/router/router_service_interface.dart
+++ b/lib/src/router/router_service_interface.dart
@@ -1,0 +1,7 @@
+import 'package:stacked/stacked.dart';
+
+abstract class RouterServiceInterface {
+  RootStackRouter get router;
+
+  RouteData get topRoute;
+}
--- a/example/navigator_example/pubspec.yaml
+++ b/example/navigator_example/pubspec.yaml
@@ -23,7 +23,7 @@ dependencies:
 
   # The following adds the Cupertino Icons font to your application.
   # Use with the CupertinoIcons class for iOS style icons.
-  cupertino_icons: ^0.1.2
+  cupertino_icons: ^1.0.2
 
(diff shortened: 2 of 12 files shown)


FilledStacks (approved):

lib/src/router/router_service_interface.dart
```diff
@@ -0,0 +1,7 @@
+import 'package:stacked/stacked.dart';
```
FilledStacks, line 1: @ferrarafer I was hoping this can go into the stacked_shared package but I see we need the RootStackRouter

FilledStacks (approved):
Besides the comment this looks good.

lib/src/router/router_service_interface.dart
ferrarafer: Yes, we need `RootStackRouter` and `RouteData` on the Interface.

PR #1087, opened 2024-03-04
feat: Added TickerProvider for ViewModel , Sample usage added

okursoftware:
https://github.com/jonataslaw/getx/blob/3de0bd3c2efe8b0537360233f4309ebba0bb038f/lib/get_state_manager/src/rx_flutter/rx_ticket_provider_mixin.dart

### DIFF
--- a/example/router_example/lib/app/app.dart
+++ b/example/router_example/lib/app/app.dart
@@ -21,6 +21,7 @@ import 'package:stacked_services/stacked_services.dart';
 import 'package:stacked_shared/stacked_shared.dart';
 import 'package:stacked_themes/stacked_themes.dart';
 
+import 'package:example/ui/ticker/ticker_view.dart';
 import 'custom_route_transition.dart';
 
 @StackedApp(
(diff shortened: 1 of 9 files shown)


example/router_example/lib/app/app.dart
```diff
@@ -39,7 +40,7 @@ import 'custom_route_transition.dart';
         page: FavoritesView,
         children: [
           MaterialRoute(page: MultipleFuturesExampleView),
-          CustomRoute(page: HistoryView),
+          //CustomRoute(page: HistoryView),
```
FilledStacks, line 43: Why is `HistoryView` removed?

FilledStacks (changes requested):
Hey, 

Can you please explain:

1. The problem you were trying to solve
2. How you're solving the problem
3. Give a concrete situation where this will be used

Thank you

eivihnd:
This seems to be the answer to one of my problems, where I need to create a TabController in the View or ViewModel.
`vsync` is needed, so either I need a StatefulWidget or something like this.

example/router_example/lib/app/app.dart
okursoftware: I got an error while building, so I deleted it.

okursoftware:
I didn't do it to solve a problem, I wanted to do it for ease of use,
 It was useful for me to use tickerprovider in viewModel class instead of making StatefullWidget unnecessarily,

PR #1117, opened 2024-08-08
Fix: Prevent swipe-to-dismiss on CupertinoRoute when blocked by PopScope

### LINKED ISSUE #1116: [bug]: CupertinoRoute do not respect PopScope when allowing swipe-to-dismiss
### Describe the bug

When using a route created by using `CupertinoRoute()` in the `@StackedApp` annotation, the iOS swipe-to-pop gesture ignores `PopScope(canPop: false)`. This means users can swipe back to pop a route that is not typically able to be popped.

### What operating system do you use?

macOS

### Information about the installed tooling

_No response_

### Steps to reproduce the issue

1. Add a `CupertinoRoute(page: MyPopScopePage)` to the `routes` list in the `@StackedApp` annotation.
2. Create a new page named `MyPopScopePage` and add a `PopScope(canPop: false)` widget to the page.
3. Navigate to the `MyPopScopePage` from another page using the `RouterService` `navigateToMyPopScopePage()` function.
4. The page will not be able to be popped via Android back buttons or Scaffold back buttons, but will be able to be popped via swiping from the left side of the screen.

### Expected behavior

Setting `PopScope(canPop: false)` should block any attempt to swipe the side of the page, similar to the behavior when using the now deprecated `WillPopScope()` and returning false in the `onWillPop` callback. 

### Screenshots

_No response_

### Additional Context

_No response_

TrentStarky:
Fixes #1116 

I have left the deprecated `if (route.hasScopedWillPopCallback) return false;` statement so this does not cause breaking changes to existing implementations, but it can be removed if wanted.

### DIFF
--- a/lib/src/router/widgets/custom_cupertino_transitions_builder.dart
+++ b/lib/src/router/widgets/custom_cupertino_transitions_builder.dart
@@ -5,11 +5,11 @@
 // The adjustments made to this code is to disable unwanted shadow
 // of routes when used as nested routes, e.g inside of a TabsRouter
 
-import 'package:flutter/cupertino.dart' show CupertinoDynamicColor;
-import 'package:flutter/foundation.dart';
 import 'dart:math';
 import 'dart:ui' show lerpDouble;
 
+import 'package:flutter/cupertino.dart' show CupertinoDynamicColor;
+import 'package:flutter/foundation.dart';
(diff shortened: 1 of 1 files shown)


FilledStacks (approved):

FilledStacks:
Thanks for the PR. This is being released

PR #1169, opened 2025-08-13
feat: Let ListenableServiceMixin implement Listenable

enwi:
Since `ListenableServiceMixin` already implements everything needed for implementing a `Listenable` it makes sense for it to implement that interface. This also opens up some other use cases for it. For example using a `ListenableServiceMixin` in a `ListenableBuilder`.


FilledStacks (approved):
This is a great suggestion, thanks.

Issue #1181, opened 2025-12-22
[bug]: Navigator2 - Navigation animation glitches on back gesture (Android only)


n0mad-d3v:
## Describe the bug
Hi guys, 
I'm having this strange issue since I migrated to navigator2: it seems that the navigation animations are doubled. Or more precisely, the forward animation is replayed before doing the backward animation when using **Back gesture on Android** (no issue on web, nor iOS). 

Here is and example (with `timeDilation = 3.0`): 

https://github.com/user-attachments/assets/c0d9b07a-df4d-4534-8ca9-01ccf05cb1f0

As I'm far from being an expert in this part of Flutter (or even any part), I tried some AI help and it find a solution, but it's using deprecated method. 
Does that make any sense to you? 

I'll do the PR anyway and try to produce a reproducible example. 
Thanks a lot for your expertise! 

Happy holidays ✨

### Issue 1: Future Already Completed

**Reproduction**:
1. Enable Navigator 2.0 in Stacked (`navigator2: true` in build.yaml)
2. Navigate to any route
3. Perform a back gesture (swipe from edge on iOS/Android)
4. Error: `Bad state: Future already completed`

**Root Cause**:
- Navigator 2.0's declarative API rebuilds page stacks during gestures
- `StackedPage.canUpdate()` returns `true` for matching pages
- The same `StackedPage` instance is reused
- `createRoute()` is called multiple times on the same instance
- Each call attempts to complete the same `_popCompleter`

**Error Stack**:
```
Unhandled Exception: Bad state: Future already completed
#0 _AsyncCompleter.complete (dart:async/future_impl.dart:97:31)
```

### Issue 2: Android Animation Glitches

**Reproduction**:
1. Run app on Android (especially Android 13+ with predictive back)
2. Perform back gestures
3. Observe: forward animation flashes before backward animation

**Root Cause**:
- `Navigator` widget was using `onDidRemovePage` callback
- Navigator 2.0 requires `onPopPage` for proper gesture handling
- Without `onPopPage`, Flutter uses default behavior causing mid-gesture rebuilds

### Fix 1: Track Route Listener State (stacked_page.dart)

Added two fields to prevent double-completion:
- `Route<T>? _currentRoute` - Tracks current route instance
- `bool _hasAttachedListener` - Prevents attaching listener multiple times

Modified `createRoute()` to:
1. Clear previous route reference (enables garbage collection)
2. Create new route
3. Only attach `popped` listener once per page lifecycle

[...]

PR #1182, opened 2025-12-24
fix(1181): Navigator2 - Navigation animation glitches on back gesture (Android only)

n0mad-d3v:
## Description

This PR fixes two issues with Navigator 2.0 integration in Stacked's router:

1. **"Future already completed" error** during back gestures
2. **Animation glitches** on Android (forward flash before backward animation)

**Reproduction**:
2. Navigate to any route

**Root Cause**:

## Deprecated warnings
It uses the deprecated member `onPopPage: (route, result)` (lib/src/router/widgets/route_navigator.dart) as I didn't find any way to fix that using `onDidRemovePage: (page)`.
Any feedback welcome on that! 

## Breaking Changes

## Checklist

- [x] Fixes are limited to the specific issues
- [x] Code follows existing style/conventions
- [x] No breaking changes to public API
- [x] Comments explain the "why" behind changes
- [x] All existing tests pass
- [ ] Added tests for the fixes (if applicable)
- [ ] Updated documentation (if needed)

## Additional Context

## Related Issues

#1181

### DIFF
--- a/lib/src/router/stacked_page.dart
+++ b/lib/src/router/stacked_page.dart
@@ -23,6 +23,11 @@ abstract class StackedPage<T> extends Page<T> {
 
   Widget get child => _child;
 
+  // Track current route and listener state to prevent double-completion
+  // Navigator 2.0 can call createRoute() multiple times on the same Page instance
+  Route<T>? _currentRoute;
+  bool _hasAttachedListener = false;
+
   StackedPage({
     required this.routeData,
     required Widget child,
@@ -60,10 +65,20 @@ abstract class StackedPage<T> extends Page<T> {
 
   @override
   Route<T> createRoute(BuildContext context) {
-    return onCreateRoute(context)
-      ..popped.then(
-        _popCompleter.complete,
-      );
+    // Clear previous route reference to allow garbage collection
+    _currentRoute = null;
+
+    // Create the new route
+    _currentRoute = onCreateRoute(context);
+
+    // Only attach listener once to prevent "Future already completed" error
+    // Navigator 2.0 can call createRoute() multiple times during back gestures
+    if (!_hasAttachedListener) {
+      _currentRoute!.popped.then(_popCompleter.complete);
+      _hasAttachedListener = true;
+    }
+
+    return _currentRoute!;
   }
 }
 
--- a/lib/src/router/widgets/route_navigator.dart
+++ b/lib/src/router/widgets/route_navigator.dart
@@ -74,10 +74,12 @@ class RouteNavigatorState extends State<RouteNavigator> {
             onDidRemovePage: (page) {
               // Critical for Navigator 2.0: Clean up page state after removal
               // The framework handles didPop and page removal automatically
-              // Use notify: false to avoid triggering rebuild during animation
+              // Defer state update to after the animation frame completes
               if (page is StackedPage) {
-                widget.router.removeRoute(page.routeData, notify: false);
-                widget.didPop?.call(page.routeData.route, null);
+                WidgetsBinding.instance.addPostFrameCallback((_) {
+                  widget.router.removeRoute(page.routeData);
+                  widget.didPop?.call(page.routeData.route, null);
+                });
               }
             },
           )
--- a/lib/src/router/widgets/route_navigator.dart
+++ b/lib/src/router/widgets/route_navigator.dart
@@ -71,16 +71,26 @@ class RouteNavigatorState extends State<RouteNavigator> {
             restorationScopeId:
                 widget.navRestorationScopeId ?? widget.router.routeData.name,
             pages: widget.router.stack,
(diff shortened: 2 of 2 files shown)


busmleon:
I assume this is the same reason why I get "future already completed" when popping a route quickly after another. Are there any plans when this will be fixed/merged?

lib/src/router/stacked_page.dart
```diff
@@ -60,10 +65,20 @@ abstract class StackedPage<T> extends Page<T> {
...
+
+    // Create the new route
+    _currentRoute = onCreateRoute(context);
+
+    // Only attach listener once to prevent "Future already completed" error
+    // Navigator 2.0 can call createRoute() multiple times during back gestures
+    if (!_hasAttachedListener) {
+      _currentRoute!.popped.then(_popCompleter.complete);
+      _hasAttachedListener = true;
+    }
```
ferrarafer: **Pop result can be lost when the page gets a second route.** The listener is attached only to the *first* route created for this page. If `createRoute()` is called again (which is the case this PR says it is handling, e.g. the `Navigator` for this `RouteNavigator` is disposed and rebuilt while the same `StackedPage` stays in `router.stack`), the first route is disposed without `didComplete` running, so its `popped` future never completes. When the user later pops the second route, nothing forwards its result, so `_popCompleter` never completes. Any `await router.push(...)` / `routingController` caller awaiting `page.popped` then hangs forever and the returned value is dropped. The old code handled this case correctly. It only threw when *both* routes completed.

Guarding the completion itself fixes both problems:

```suggestion
    _currentRoute!.popped.then((value) {
      if (!_popCompleter.isCompleted) _popCompleter.complete(value);
    });
```
(then `_hasAttachedListener` and `_currentRoute` are no longer needed. Keeping `_currentRoute` on the page also keeps the disposed route alive for as long as the page is in the stack, which is the opposite of what the "allow garbage collection" comment says.)

lib/src/router/widgets/route_navigator.dart
```diff
@@ -71,12 +71,26 @@ class RouteNavigatorState extends State<RouteNavigator> {
             restorationScopeId:
                 widget.navRestorationScopeId ?? widget.router.routeData.name,
             pages: widget.router.stack,
-            onDidRemovePage: (page) {
-              if (page is StackedPage) {
-                var routeData = page.routeData;
-                widget.router.removeRoute(routeData);
-                widget.didPop?.call(routeData.route, null);
+            // ignore: deprecated_member_use
+            onPopPage: (route, result) {
```
ferrarafer: **Regression: imperative page removals no longer sync the router stack.** With `onPopPage` set instead of `onDidRemovePage`, Flutter only calls back for `Navigator.pop` / `maybePop` / `popUntil`. Every other imperative removal of a page-backed route (`Navigator.removeRoute`, `removeRouteBelow`, `pushReplacement`, `pushAndRemoveUntil`, `replace`/`replaceRouteBelow`) goes through `_RouteEntry.complete(..., imperativeRemoval: true)`, which only calls `widget.onDidRemovePage` (see `navigator.dart` `handlePop`/`handlePush`). The old code handled those through `onDidRemovePage`.

Scenario: app code calls `Navigator.of(context).pushAndRemoveUntil(...)` or `removeRoute(route)` on a `StackedPage` route. The route gets disposed, but its `StackedPage` stays in `router._pages`. `removeRoute`/`notifyAll` never run, so the URL, guards and `didPop` go stale. On the next `notifyAll` rebuild the Navigator finds a page with no entry and calls `createRoute` on the same `StackedPage` again, so the removed screen comes back. That is also the "createRoute called twice on the same page" case the completer guard in `stacked_page.dart` now hides silently.

lib/src/router/stacked_page.dart
```diff
@@ -50,20 +48,20 @@ abstract class StackedPage<T> extends Page<T> {
...
   Route<T> createRoute(BuildContext context) {
+    // Navigator 2.0 can call createRoute() multiple times on the same page
+    // (e.g. when the Navigator is rebuilt), so only the first popped route
+    // completes the page's completer.
     return onCreateRoute(context)
-      ..popped.then(
-        _popCompleter.complete,
-      );
+      ..popped.then((value) {
+        if (!_popCompleter.isCompleted) _popCompleter.complete(value);
```
ferrarafer: This guard hides the symptom rather than fixing it. When `createRoute` runs a second time on the same `StackedPage` (for example the page resurrected after an imperative removal, as described on `route_navigator.dart`), the page completer has already completed with the first route's result, usually `null` for a removal. Whatever the live second route pops with is then dropped silently, so a caller awaiting `router.push<T>()` got `null` early and never sees the real result. Before this change the case showed up as a `StateError`. Now it loses data without any signal. Consider at least an assert or debug log when the completer is already completed.

PR #1192, opened 2026-09-17
feat(viewmodels): add safeData accessor to DataStateHelper

### LINKED ISSUE #1118: [feature]: Asserting values are present when viewmodel.dataReady is true
### Is your feature request related to a problem? Please describe.

When using a FutureViewModel, !viewModel.dataReady is a nice helper to indicate whether the data has been loaded. When it is true, one can be sure that the values are there. However, I still need to use ! on the viewModel.data in this case, despite being sure that the data is there. This is error prone because I might use the ! in instances where the data is indeed not there.

### Describe the solution you would like

I image some kind of assertion when viewModel.dataReady is true such that when viewModel.data is used under this scope, the non-null return type can be used.
For now i have written the following extension function such that i can use viewModel.safeData, but i think it might be possible to use viewModel.data, and deduce the information that the data is loaded.

```
import 'package:stacked/stacked.dart';

extension SafeDataAccess<T> on DataStateHelper<T> {
  /// Asserts that data is non-null if dataReady is true and returns the data.
  T get safeData {
    if (!dataReady) {
      throw StateError('Data is not ready to be accessed');
    }
    return data as T; // Since dataReady is true, data is guaranteed to be non-null.
  }
}
```

### Additional Context

image
image

Eimen2018:
## Summary

Adds a `safeData` getter to `DataStateHelper<T>` (used by `FutureViewModel` and `StreamViewModel`).

`data` is `T?` because it starts unset and is cleared on error, so Dart cannot promote it to non-null even right after a `dataReady` check. Callers end up writing `data!` everywhere. `safeData` performs the `dataReady` check itself and returns a non-null `T`, or throws a descriptive `StateError` pointing the caller at `dataReady` / `isBusy` / `hasError`.

```dart
if (viewModel.dataReady) {
  final value = viewModel.safeData; // no `!` needed
}
```

This is the accessor agreed on in the issue thread; the requested PR was never submitted.

## Changes

- `lib/src/view_models/helpers/data_state_helper.dart`: new `T get safeData` with dartdoc. Existing `data` and `dataReady` are untouched.
- `test/data_state_helper_test.dart`: 8 tests covering ready, never set, busy, error, error message content, and nullable `T` behaviour.

## Notes

- Purely additive, no signature changes, no new dependencies.
- `MultipleFutureViewModel` / `MultipleStreamViewModel` were left alone: their `dataMap` is `Map<K, dynamic>?` on a private base class, so a keyed `safeData` would give no type-safety benefit.
- Nullable `T` whose legitimate value is `null` is never `dataReady` today; `safeData` keeps that semantic and documents it.

## Verification

- `flutter test`: 101 passed

Closes #1118


Eimen2018:
Superseded by #1196, which is the same change opened from a branch in this repository instead of a fork so that CI can post the coverage report.

PR #1193, opened 2026-09-17
fix(router): guard StackedPage pop completer and eagerly remove popped page

### LINKED ISSUE #1181: [bug]: Navigator2 - Navigation animation glitches on back gesture (Android only)

Eimen2018:
## Summary

Fixes #1181: on Android back gestures with the Navigator 2 router, users hit `Bad state: Future already completed` and see the forward transition flash before the back animation.

## Root cause

`StackedPage` holds a single `_popCompleter` per page instance and `createRoute()` chained `popped.then(_popCompleter.complete)` on every call. Flutter can call `createRoute()` more than once for the same `Page` instance. Between a pop and the async `onDidRemovePage` callback, the Navigator may rebuild while `router.stack` still contains the popped page. Seeing a page with no live Route behind it, the Navigator creates a phantom second Route. That phantom route runs an enter animation (the visual flash) and, when it eventually pops, completes the same completer a second time (the crash).

## Fix

`lib/src/router/stacked_page.dart`:

1. Track the latest route created for the page. Only the latest route's pop result is forwarded, and only if the completer is not already completed. A stale route can no longer complete the completer or leak its result.
2. When the latest route pops, eagerly call `router.removeRoute(routeData, notify: true)` so the Navigator receives an updated `pages` list before it can recreate the route. `_removeRoute` is idempotent (`lastIndexWhere` then `removeAt` only if found), so the later `onDidRemovePage` call is a harmless no-op. This runs inside a `Future.then` microtask, after the current frame has unwound, so it does not notify during a build.

The eager-removal approach is based on #1188 by @Vinsho. Compared to that PR, this adds the latest-route check, which prevents a stale route's pop from forwarding a wrong result or removing the page prematurely.

## Relation to #1182

#1182 switches `RouteNavigator` to the deprecated `onPopPage`. This PR keeps `onDidRemovePage`. #1182's completer guard also attaches the listener only once, so a second `createRoute` call returns a route whose pop is never observed and `popped` would never resolve. That approach is not adopted here.

## Tests

`test/router/stacked_page_test.dart` (new, 5 tests): single pop forwards the result; `createRoute` called twice does not throw and forwards only the latest route's result; stale route pop is ignored; popping removes the page from the router without waiting for `onDidRemovePage`; with two routes only the latest triggers exactly one `removeRoute`. The double-completion tests fail on `master` without the fix.

## Verification

- `flutter test`: 98 passed, 1 pre-existing skip
- No public API change


Eimen2018:
Superseded by #1197, which is the same change opened from a branch in this repository instead of a fork so that CI can post the coverage report.

PR #1167, opened 2025-08-06
feat: modernize codebase patterns and dependency update

Eimen2018:
## Summary
  This PR modernizes the Stacked codebase to align with current Flutter and Dart best practices, including dependency updates, constructor patterns, and
   code generation improvements.

  ## Changes Made

  ### Dependency Updates
  - Update core dependencies to latest compatible versions:
    - `get_it`: ^8.0.2 → ^8.2.0
    - `meta`: ^1.15.0 → ^1.16.0
    - `provider`: ^6.1.2 → ^6.1.5
    - `collection`: ^1.18.0 → ^1.19.1
    - `path`: ^1.9.0 → ^1.9.1
  - Add `web: ^1.1.1` for Flutter 3.0+ compatibility

  ### Code Modernization
  - **Library Syntax**: Updated to modern unnamed library syntax (`library;`)
  - **Constructor Pattern**: Migrated 59+ files from `Key? key` to `super.key` pattern
  - **Export Organization**: Improved alphabetical ordering of library exports

  ### Generated Code Updates
  - Regenerated router and app configuration files
  - Updated form generation code
  - Improved router navigation implementation
  - Updated example applications with latest patterns

  ## Files Changed
  - **Core Library**: 8 files in `lib/src/`
  - **Examples**: 59 files across navigator and router examples
  - **Generated**: App routers, locators, and form files

  ## Testing
  - [x] All existing tests pass
  - [ ] Example applications build and run successfully
  - [ ] No breaking changes in public API

  ## Migration Guide
  For users extending Stacked widgets, update constructor patterns:
  ```dart
  // Before
  MyWidget({Key? key}) : super(key: key);

  // After
  MyWidget({super.key});

PR #1195, opened 2026-09-17
ci: only post coverage comment for same-repo pull requests

Eimen2018:
## Summary

Every pull request opened from a fork currently fails CI at the "Report code coverage" step with `Resource not accessible by integration`, even when `dart fix`, analysis and `flutter test` all pass. Recent examples: #1167, #1169, #1192, #1193.

## Cause

`zgosalvez/github-actions-report-lcov` posts a coverage comment on the PR using the `github-token` input. For `pull_request` events from forks, GitHub hands the workflow a read-only `GITHUB_TOKEN`, so the comment API call is rejected and the step fails.

## Fix

The action documents `github-token` as optional and skips commenting when it is empty. This change passes the token only when the PR head is in this repository:

```yaml
github-token: ${{ github.event.pull_request.head.repo.full_name == github.repository && secrets.GITHUB_TOKEN || '' }}
```

Same-repo branches keep the coverage comment. Fork PRs still run coverage and upload the HTML artifact, they just do not comment.

Once merged, re-running checks on the open fork PRs will pick this up since the run uses the merge commit with `master`.


Eimen2018:
Closing for now. The PRs this was blocking have been reopened from branches in this repository, so they no longer hit the read-only token limitation. The underlying problem still applies to pull requests from forks: the coverage comment step fails with "Resource not accessible by integration" even when lint and tests pass. This change can be reopened if that becomes a nuisance for outside contributors.

PR #1198, opened 2026-09-18
fix(router): avoid notifying listeners during build on page removal

### LINKED ISSUE #1184: [bug]: setState() called during build - Exception on Browser Back Navigation
### Describe the bug

When using the browser's built-in back navigation controls in a Stacked web application, the following exception is thrown:

```
══╡ EXCEPTION CAUGHT BY FOUNDATION LIBRARY ╞══
The following assertion was thrown while dispatching notifications for StackedRouterWeb:
setState() or markNeedsBuild() called during build.
```

This issue occurs using the default Stacked web template (unmodified code generated by the CLI).

### What operating system do you use?

macOS

### Information about the installed tooling

```
Flutter version: 3.35.4
Stacked CLI version: 1.15.4

stacked: 3.4.0
stacked_services: 1.1.0
stacked_generator: 1.3.3

Platform: Web (running on Google Chrome)
```

The issue has persisted even when updating `stacked`, `stacked_services`, and `stacked_generator` to their latest versions. The above versions are the default versions used after running `stacked create app -t web`.

### Steps to reproduce the issue

1. Create a new Stacked web app: `stacked create app example -t web`
2. Run app in Google Chrome.
3. Navigate from the `HomeView` to the `UnknownView` (or any view) using the `RouterService`, for example: 

```
class HomeViewDesktop extends ViewModelWidget<HomeViewModel> {
  const HomeViewDesktop({super.key});

  @override
  Widget build(BuildContext context, HomeViewModel viewModel) {
    return Scaffold(
      body: Center(
          child: ElevatedButton(
              onPressed: () {
                locator<RouterService>().navigateToUnknownView();
              },
              child: const Text('Go to Unknown view'))),
    );
  }
}
```

4. Navigate back using the browser’s back navigation control.
5. Exception gets thrown.

### Expected behavior

Browser back navigation should work without throwing exceptions, similar to programmatic navigation using `locator<RouterService>().back();` - which works without issues.

### Screenshots

_No response_

### Additional Context

<details>

<summary>Here is the complete error:</summary>

```
══╡ EXCEPTION CAUGHT BY FOUNDATION LIBRARY ╞════════════════════════════════════════════════════════
The following assertion was thrown while dispatching notifications for StackedRouterWeb:
setState() or markNeedsBuild() called during build.
This _RootRouter widget cannot be marked as needing to build because the framework is already in the
process of building widgets. A widget can be marked as needing to be built during the build phase
only if one of its ancestors is currently building. This exception is allowed because the framework
builds parent widgets before children, which means a dirty descendant will always be built.
Otherwise, the framework might not visit this widget during this build phase.
The widget on which setState() or markNeedsBuild() was called was:
  _RootRouter
The widget which was currently being built when the offending call was made was:
  RouteNavigator

When the exception was thrown, this was the stack:
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/errors.dart 274:3       throw_
errors.dart:274
package:flutter/src/widgets/framework.dart 5279:9                                 <fn>
framework.dart:5279
package:flutter/src/widgets/framework.dart 5290:14                                markNeedsBuild
framework.dart:5290
package:flutter/src/widgets/framework.dart 1219:5                                 setState
framework.dart:1219
package:stacked/src/router/controller/nested_router_delegate.dart 176:7           [_handleRebuild]
nested_router_delegate.dart:176
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart 118:77  tear
operations.dart:118
package:flutter/src/foundation/change_notifier.dart 435:24                        notifyListeners
change_notifier.dart:435
package:stacked/src/router/controller/routing_controller.dart 79:5                notifyAll
routing_controller.dart:79
package:stacked/src/router/controller/routing_controller.dart 884:7               [_removeRoute]
routing_controller.dart:884
package:stacked/src/router/controller/routing_controller.dart 866:5               removeRoute
routing_controller.dart:866
package:stacked/src/router/widgets/route_navigator.dart 77:17                     <fn>
route_navigator.dart:77
package:flutter/src/widgets/navigator.dart 3307:41                                handlePop
navigator.dart:3307
package:flutter/src/widgets/navigator.dart 4450:21                                [_flushHistoryUpdates]
navigator.dart:4450
package:flutter/src/widgets/navigator.dart 4378:5                                 [_updatePages]
navigator.dart:4378
package:flutter/src/widgets/navigator.dart 4013:7                                 didUpdateWidget
navigator.dart:4013
package:flutter/src/widgets/framework.dart 5893:55                                update
framework.dart:5893
package:flutter/src/widgets/framework.dart 3982:14                                updateChild
framework.dart:3982
package:flutter/src/widgets/framework.dart 5747:16                                performRebuild
framework.dart:5747
package:flutter/src/widgets/framework.dart 5884:11                                performRebuild
framework.dart:5884
package:flutter/src/widgets/framework.dart 5435:7                                 rebuild
framework.dart:5435
package:flutter/src/widgets/framework.dart 5909:5                                 update
framework.dart:5909
package:flutter/src/widgets/framework.dart 3982:14                                updateChild
framework.dart:3982
package:flutter/src/widgets/framework.dart 5747:16                                performRebuild
framework.dart:5747
package:flutter/src/widgets/framework.dart 5435:7                                 rebuild
framework.dart:5435
package:flutter/src/widgets/framework.dart 6051:5                                 update
framework.dart:6051
package:flutter/src/widgets/framework.dart 3982:14                                updateChild
framework.dart:3982
package:flutter/src/widgets/framework.dart 5747:16                                performRebuild
framework.dart:5747
package:flutter/src/widgets/framework.dart 5435:7                                 rebuild
framework.dart:5435
package:flutter/src/widgets/framework.dart 6051:5                                 update
framework.dart:6051
package:flutter/src/widgets/framework.dart 3982:14                                updateChild
framework.dart:3982
package:flutter/src/widgets/framework.dart 5747:16                                performRebuild
framework.dart:5747
package:flutter/src/widgets/framework.dart 5884:11                                performRebuild
framework.dart:5884
package:flutter/src/widgets/framework.dart 5435:7                                 rebuild
framework.dart:5435
package:flutter/src/widgets/framework.dart 5909:5                                 update
framework.dart:5909
package:flutter/src/widgets/framework.dart 3982:14                                updateChild
framework.dart:3982
package:flutter/src/widgets/framework.dart 5747:16                                performRebuild
framework.dart:5747
package:flutter/src/widgets/framework.dart 5435:7                                 rebuild
framework.dart:5435
package:flutter/src/widgets/framework.dart 5797:5                                 update
framework.dart:5797
package:flutter/src/widgets/framework.dart 3982:14                                updateChild
framework.dart:3982
package:flutter/src/widgets/framework.dart 5747:16                                performRebuild
framework.dart:5747
package:flutter/src/widgets/framework.dart 5435:7                                 rebuild
framework.dart:5435
package:flutter/src/widgets/framework.dart 6051:5                                 update
framework.dart:6051
package:flutter/src/widgets/framework.dart 3982:14                                updateChild
framework.dart:3982
package:flutter/src/widgets/framework.dart 5747:16                                performRebuild
framework.dart:5747
package:flutter/src/widgets/framework.dart 5435:7                                 rebuild
framework.dart:5435
package:flutter/src/widgets/framework.dart 6051:5                                 update
framework.dart:6051
package:flutter/src/widgets/framework.dart 3982:14                                updateChild
framework.dart:3982
package:flutter/src/widgets/framework.dart 5747:16                                performRebuild
framework.dart:5747
package:flutter/src/widgets/framework.dart 5884:11                                performRebuild
framework.dart:5884
package:flutter/src/widgets/framework.dart 5435:7                                 rebuild
framework.dart:5435
package:flutter/src/widgets/framework.dart 2695:14                                [_tryRebuild]
framework.dart:2695
package:flutter/src/widgets/framework.dart 2752:11                                [_flushDirtyElements]
framework.dart:2752
package:flutter/src/widgets/framework.dart 3056:17                                buildScope
framework.dart:3056
package:flutter/src/widgets/layout_builder.dart 271:5                             [_rebuildWithConstraints]
layout_builder.dart:271
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart 118:77  tear
operations.dart:118
package:flutter/src/widgets/layout_builder.dart 334:28                            layoutCallback
layout_builder.dart:334
package:flutter/src/rendering/object.dart 4156:33                                 <fn>
object.dart:4156
package:flutter/src/rendering/object.dart 2881:9                                  <fn>
object.dart:2881
package:flutter/src/rendering/object.dart 1206:7                                  [_enableMutationsToDirtySubtrees]
object.dart:1206
package:flutter/src/rendering/object.dart 2880:7                                  invokeLayoutCallback
object.dart:2880
package:flutter/src/rendering/object.dart 4156:5                                  runLayoutCallback
object.dart:4156
package:flutter/src/widgets/layout_builder.dart 448:5                             performLayout
layout_builder.dart:448
package:flutter/src/rendering/object.dart 2610:7                                  [_layoutWithoutResize]
object.dart:2610
package:flutter/src/rendering/object.dart 1157:17                                 flushLayout
object.dart:1157
package:flutter/src/rendering/object.dart 1170:14                                 flushLayout
object.dart:1170
package:flutter/src/rendering/binding.dart 629:5                                  drawFrame
binding.dart:629
package:flutter/src/widgets/binding.dart 1261:13                                  drawFrame
binding.dart:1261
package:flutter/src/rendering/binding.dart 495:5                                  [_handlePersistentFrameCallback]
binding.dart:495
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart 118:77  tear
operations.dart:118
package:flutter/src/scheduler/binding.dart 1434:7                                 [_invokeFrameCallback]
binding.dart:1434
package:flutter/src/scheduler/binding.dart 1347:9                                 handleDrawFrame
binding.dart:1347
package:flutter/src/scheduler/binding.dart 1200:5                                 [_handleDrawFrame]
binding.dart:1200
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart 118:77  tear
operations.dart:118
lib/_engine/engine/platform_dispatcher.dart 1522:5                                invoke
platform_dispatcher.dart:1522
lib/_engine/engine/platform_dispatcher.dart 265:5                                 invokeOnDrawFrame
platform_dispatcher.dart:265
lib/_engine/engine/frame_service.dart 192:32                                      [_renderFrame]
frame_service.dart:192
lib/_engine/engine/frame_service.dart 99:9                                        <fn>
frame_service.dart:99
dart-sdk/lib/async/zone.dart 1849:54                                              runUnary
zone.dart:1849
dart-sdk/lib/async/zone.dart 1804:26                                              <fn>
zone.dart:1804
dart-sdk/lib/_internal/js_dev_runtime/patch/js_allow_interop_patch.dart 224:27    _callDartFunctionFast1
js_allow_interop_patch.dart:224

The StackedRouterWeb sending notification was:
  Root Router
════════════════════════════════════════════════════════════════════════════════════════════════════

```
</details>

**Thank you to all the contributors for their hard work on this brilliant package!**

Eimen2018:
## Summary

Fixes #1184: browser back navigation on a Stacked web app (default `stacked create app -t web` template) throws `setState() or markNeedsBuild() called during build` from `_RootRouter`.

## Root cause

Matched against the reported stack trace:

1. `Navigator.onDidRemovePage` in `route_navigator.dart` calls `router.removeRoute(routeData)`.
2. `StackRouter._removeRoute` called `notifyAll(forceUrlRebuild: true)` unconditionally, even when the page was already gone from `_pages`.
3. `notifyAll` triggers `_RootRouterState._handleRebuild`, which calls `setState()`.

When the removal is a side effect of the Navigator updating while the router widget's own rebuild is still in progress, that synchronous `setState()` lands during build and throws.

## Fix

`lib/src/router/controller/routing_controller.dart`:

- **No-op removal no longer notifies.** When the page is not in `_pages`, nothing is removed, shared path data is left alone and listeners are not notified. Redirect-guard cleanup and `_removeTopRouterOf` still run, as on `master`, so pages taken out directly by `removeWhere`, `_reset` or `updateDeclarativeRoutes` still get their guard listeners and child routers cleaned up.
- **`_removeRedirectGuard` is null-safe.** It does nothing when the guard has no registered listener.
- **Build-phase guard.** When a page is removed during `SchedulerPhase.persistentCallbacks`, the notification is deferred with `addPostFrameCallback`. In every other phase it stays synchronous.
- **Dispose guard.** `StackRouter` tracks `_disposed`, and the deferred callback returns early on a disposed router, so a nested router unmounted later in the same frame is not notified.

## Tests

- `browser_back_navigation_test.dart`: reproduces the reported exception on `master`, passes with the fix. Plus an end-to-end `handlePopRoute()` check.
- `router_cleanup_test.dart`: a route with a `RedirectGuard` is removed via `removeWhere`, then the Navigator's follow-up `removeRoute` hits the not-found path. Asserts the guard listener is detached and a later `reevaluate` does not re-push. Failed with the earlier early return, passes now.
- `nested_router_dispose_test.dart`: a nested router schedules the deferred notify and is disposed in the same frame. Threw "used after being disposed" before the guard, passes now.

## Verification

- `flutter test`: 97 passed, 1 pre-existing skip
- No public API change

Note: #1197 also touches `_removeRoute`. Whichever merges second will need a small rebase.


lib/src/router/controller/routing_controller.dart
```diff
@@ -868,9 +869,11 @@ abstract class StackRouter extends RoutingController {
 
   void _removeRoute(RouteMatch route, {bool notify = true}) {
     var pageIndex = _pages.lastIndexWhere((p) => p.routeKey == route.key);
-    if (pageIndex != -1) {
-      _pages.removeAt(pageIndex);
+    if (pageIndex == -1) {
+      // Already removed by another path; nothing to notify.
+      return;
```
ferrarafer: **Early return skips the redirect-guard cleanup and child-router cleanup when the page is already gone from `_pages`.**

`removeWhere` (and `_reset` / `updateDeclarativeRoutes`) take pages out of `_pages` directly. They never call `_removeRedirectGuard` or `_removeTopRouterOf`. Before this PR, when the Navigator then dropped that page, `onDidRemovePage` -> `removeRoute` hit `pageIndex == -1` but still ran the guard loop and `_removeTopRouterOf`, so the listener got cleaned up there. With this early return that never happens.

Scenario: a route guarded by a `RedirectGuard` is removed with `router.removeWhere(...)`. Its listener stays in `_redirectGuardsListeners`. The next time the guard calls `notifyListeners()` (for example on an auth state change), `_reevaluate` runs on this router. No guarded route is left in the stack, so it goes straight to `_strategy.reevaluate(...)`, and that can re-push routes or `removeAllAndPush` for a page the user already left.

Suggestion: skip only the `_pages.removeAt` and the notification when `pageIndex == -1`. Keep the cleanup, but make it null-safe. For example, `_removeRedirectGuard` should do nothing if `_redirectGuardsListeners[guard]` is null.

lib/src/router/controller/routing_controller.dart
```diff
@@ -881,6 +884,19 @@ abstract class StackRouter extends RoutingController {
...
+    }
+  }
+
+  // Defers the notification when called mid-build to avoid setState during
+  // build.
+  void _notifyRouteRemoved() {
+    if (SchedulerBinding.instance.schedulerPhase ==
+        SchedulerPhase.persistentCallbacks) {
+      SchedulerBinding.instance.addPostFrameCallback((_) {
+        notifyAll(forceUrlRebuild: true);
```
ferrarafer, line 907: **The deferred `notifyAll` can run on a router that has already been disposed.**

The post-frame closure holds onto `this` and never checks whether the controller is still alive. Nested routers dispose their `StackRouter` in `State.dispose` (`nested_router.dart`, `nested_router_delegate.dart`). Unmounting happens during `finalizeTree`, which is still in the same `persistentCallbacks` phase that scheduled the callback.

Scenario: a nested router gets `onDidRemovePage` during the build phase, and its widget is unmounted later in that same frame. The post-frame callback then calls `notifyListeners()` on the disposed `ChangeNotifier`. In debug builds this throws "A StackRouter was used after being disposed". It also still triggers `root.notifyListeners()` and `navigationHistory.rebuildUrl()` from a stale router, which can overwrite the URL with the torn-down state.

Suggestion: track disposal (for example a `_disposed` flag set in `dispose`) and return early from the callback when it is set.

ferrarafer (changes requested):

lib/src/router/controller/routing_controller.dart
```diff
@@ -878,9 +886,27 @@ abstract class StackRouter extends RoutingController {
...
     _removeTopRouterOf(route.key);
-    if (notify) {
+    if (removed && notify) {
+      _notifyRouteRemoved();
+    }
+  }
+
+  // Defers the notification when called mid-build to avoid setState during
+  // build.
+  void _notifyRouteRemoved() {
```
ferrarafer, line 902: **[medium] The build-phase guard is applied only to `_removeRoute`, so the reported crash is still reachable through the sibling removal paths.**

`notifyAll(forceUrlRebuild: true)` is also called synchronously from `removeWhere` (~line 1089) and `_removeUntil` (~line 1075), and `notifyAll()` from `_addEntry` (~line 1181). None of them go through `_notifyRouteRemoved`, so they still call `setState()`/`markNeedsBuild()` during build.

I verified this on this branch: a descendant widget whose `build` calls `router.removeWhere((r) => r.name == 'SecondRoute')` (the same shape as the first new test, which uses `removeRoute`) still throws:

```
The following assertion was thrown while dispatching notifications for _R:
setState() or markNeedsBuild() called during build.
The widget on which setState() or markNeedsBuild() was called was: _RootRouter
...
The following assertion was thrown while dispatching notifications for NestedRouterDelegate:
setState() or markNeedsBuild() called during build.
The widget on which setState() or markNeedsBuild() was called was: Router<Object>
```

Note this also shows the second setState source (`NestedRouterDelegate._handleRebuild` via `navigationHistory.rebuildUrl()`), which `_notifyRouteRemoved` happens to cover for the `_removeRoute` path but not elsewhere.

Since the helper already exists, consider making it a general `_notifySafely({bool forceUrlRebuild})` and routing `removeWhere`, `_removeUntil` and `_addEntry` through it too — otherwise the fix is specific to the one stack frame in #1184 rather than to the class of bug.

test/router/browser_back_navigation_test.dart
```diff
@@ -0,0 +1,145 @@
...
+      );
+      await tester.pumpAndSettle();
+      expect(find.text('Second'), findsOneWidget);
+
+      await tester.binding.handlePopRoute();
+
+      for (var i = 0; i < 10; i++) {
+        await tester.pump(const Duration(milliseconds: 50));
+        expect(tester.takeException(), isNull);
+      }
```
ferrarafer, line 142: **[low] This second test passes on `master` too, so it does not guard the fix — and it never asserts that the back navigation actually did anything.**

I ran the three new test files against the pre-fix `routing_controller.dart`: `browser_back_navigation_test.dart` test #1 and `nested_router_dispose_test.dart` fail (good reproductions), but this `handlePopRoute()` test passes both before and after. So the "real system/browser back navigation" path as modelled here never actually hit the reported exception.

It is also vacuous in a second way: the only assertion is `takeException() == null`. If `handlePopRoute()` silently did nothing, the test would still pass. Instrumenting it shows the pop does happen (`router.stack.length == 1`, `currentUrl == '/'`), so those are cheap assertions to add:

```suggestion
      await tester.binding.handlePopRoute();

      for (var i = 0; i < 10; i++) {
        await tester.pump(const Duration(milliseconds: 50));
        expect(tester.takeException(), isNull);
      }

      expect(router.stack, hasLength(1));
      expect(find.text('Home'), findsOneWidget);
      expect(router.currentUrl, '/');
```

Worth flagging in the PR description too, since it is presented as the end-to-end check for #1184 — the only test that actually reproduces the reported exception is test #1, which triggers the removal from a descendant's `build` rather than from a genuine browser back.

lib/src/router/controller/routing_controller.dart
```diff
@@ -878,9 +886,27 @@ abstract class StackRouter extends RoutingController {
...
+    if (SchedulerBinding.instance.schedulerPhase ==
+        SchedulerPhase.persistentCallbacks) {
+      SchedulerBinding.instance.addPostFrameCallback((_) {
+        if (_disposed) return;
```
ferrarafer: When the deferred callback sees `_disposed`, it returns without doing anything. That also skips the other two things `notifyAll` does for a nested router: `root.notifyListeners()` and `navigationHistory.rebuildUrl()`. Neither `NestedRouter`'s dispose nor `removeChildController` rebuilds the URL.

Here is how it goes wrong. A nested router removes its top page during build, and its host widget is unmounted in the same frame while the parent page stays on the stack. This is the setup in `nested_router_dispose_test.dart`, where `_NestedHost` switches to `SizedBox.shrink()`. The root is never notified, and the browser URL still shows the removed nested path (`/parent/inner-home`) until some unrelated navigation happens.

One option is to keep the guard on this router's own listeners but still notify the root and rebuild the URL when `!isRoot`. The other is to say in the comment that dropping them is intended.

ferrarafer: Fixed in 111f6964. When the router is disposed and is not the root, the deferred callback now still calls `root.notifyListeners()` and `navigationHistory.rebuildUrl()`; only its own `notifyListeners()` is skipped, since that is the call that would throw on a disposed `ChangeNotifier`.

`nested_router_dispose_test.dart` now asserts the url after the frame settles. Without the fix it fails with `Expected: '/parent'  Actual: '/inner-home'`.

Issue #1105, opened 2024-06-01
[feature]: WebAssembly (Wasm) Support


BrunoMiguens:
### Is your feature request related to a problem? Please describe.

Currently `web_navigation_history.dart` uses `dart:html` which is nt supported by Wasm and should be migrated to `package:web` - [Reference](https://docs.flutter.dev/platform-integration/web/wasm#js-interop-wasm)

### Describe the solution you would like

Replace the usage of `dart:html` to use the new web library and support the latest stable Wasm.
As far as I could check, it should be a matter of changing the `_history.state` usage, however, I'm not sure if that is all.

### Additional Context

Flutter 3.22

#### Support for Wasm is now stable!
WebAssembly support for Flutter web is available on the Flutter [stable](https://github.com/flutter/flutter/wiki/flutter-build-release-channels#stable) channel.

#### Dart's next-gen web interop is now stable!
Migrate your packages to [package:web](https://pub.dev/packages/web) and [dart:js_interop](https://api.dart.dev/stable/dart-js_interop) to make them compatible with Wasm. Read the [Requires JS-interop](https://docs.flutter.dev/platform-integration/web/wasm#js-interop-wasm) section to learn more.

FilledStacks:
Thanks @BrunoMiguens for filing this, it's a very requested feature that we'll be adding in. 

We're working on some web apps for clients that also want WASM so it'll be in the up next column soon.

BrunoMiguens:
Thanks, that's is great to know as it looks like there's not a lot of work, from a very high level perspective.

FilledStacks:
Yup, lots to do there. 

I'll be dedicating a few days to it soon.

DevGranGerda:
@FilledStacks 
Upgrading Get dependency to 4.6.6 solves issues with Stacked Services, where dart:html is removed. My present hiccup when trying to build with --wasm flag.

FilledStacks:
Awesome! Thanks @Schefferbird we'll do that and push and update to the services now

FilledStacks:
Try version 1.5.1 of Stacked services and see if you can compile to WASM @BrunoMiguens

SHuygh:
@FilledStacks I have tried compiling for wasm with version 1.5.1 and still getting errors on the Get dependency.

Apparently Get 4.6.6 still uses dart:html, their release candidate no longer has this issue, but this is not stable yet and is not compatible with stacked services atm.

Eimen2018:
[...]

PR #1199, opened 2026-09-19
fix: drop universal_io so the package is WASM compatible

Eimen2018:
## Summary

Makes `stacked` WASM compatible on pub.dev. Refs #1105.

## Problem

pub.dev marks `stacked` 3.5.0 as not WASM compatible, and `stacked_services` inherits that because it imports `stacked`. The reason pub.dev gives:

`package:stacked/stacked.dart` → `router_utils.dart` → `package:universal_io/io.dart` → ... → `internet_address.dart` → `dart:io`

`universal_io` was only used for one check, `Platform.isIOS || Platform.isMacOS` in `buildAdaptivePageRoute`, and that check already sits behind a `kIsWeb` guard. The latest `universal_io` is itself not WASM ready, so upgrading it does not help.

## Change

- `router_utils.dart`: use `defaultTargetPlatform` from `flutter/foundation` instead of `Platform`.
- `pubspec.yaml`: remove `universal_io`.

Behaviour is the same on every platform. A side benefit is that the check now respects `debugDefaultTargetPlatformOverride` in tests.

## Verification

- `pana` (the analyser pub.dev uses) on this branch: **`is:wasm-ready`**, platform support 20/20. On `master` it is not WASM ready.
- Built a test app with `flutter build web --wasm` against this branch and ran it in a browser. Confirmed it ran the `.wasm` module with no JS fallback, `buildAdaptivePageRoute` worked, and the console was clean.
- `flutter test`: 93 passed. `dart analyze`: no issues in the changed file.


ferrarafer (approved):
LGTM

